PR #7, opened 2018-07-24
Support graph types on the driver level though value handlers

ali-ince:
Previously graph types were deserialised on the go-connector level and did not support `Path` values to be consumed through the driver.

With this PR (and improvements on `neo4j-go-connector`), graph types are now deserialised on the driver level and added support for other future types.


ali-ince:
[Add extensible Bolt Value serializer/deserializer at the driver level](https://trello.com/c/FTuSJR7X/3-add-extensible-bolt-value-serializer-deserializer-at-the-driver-level)

PR #96, opened 2020-05-07
Clearified minimum viable snippet

2hdddg:
Needs a different configuration when running against v4 db.
Removed var section as well.


martin-neotech (approved):
Perfect, this explicit config for the specific Neo4j server version will save the user from the confusion of the encryption changes of default setting of the Neo4j server.

PR #97, opened 2020-05-11
Better error message on 3.5 vs 4.0 TLS change

2hdddg:
Neo4j server 3.5 generated certificates automatically, 4.0 does not and
has therefore TLS disabled by default. This causes confusion upon
upgrade and this improves that error message.

### DIFF
--- a/neo4j/connector.go
+++ b/neo4j/connector.go
@@ -24,6 +24,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


neo4j/connector.go
```diff
@@ -139,6 +140,9 @@ func (c *connector) wrapInTls(target string, rawConn net.Conn) (net.Conn, error)
 	tlsConn := tls.Client(rawConn, &conf)
 	err = tlsConn.Handshake()
 	if err != nil {
+		if err == io.EOF {
+			err = errors.New("Remote end closed the connection, check that TLS is enabled on the server")
```
martin-neotech, line 144: Either it is that TLS should be enabled on the server or that the client config should be set to not to use an encrypted connection. But yes it is nice to get an error message that hints that it is probably something to do with TLS settings.

2hdddg: Yes, I assume the client is right ;-)

PR #122, opened 2020-07-20
added default port 7687

martin-neotech:


2hdddg (changes requested):
Please run gofmt to fix indentation (or do it manually) and add a test to prove that it works.

neo4j/driver_test.go
```diff
@@ -129,6 +129,17 @@ func TestDriverN(t0 *testing.T) {
...
 			assertError(t, err)
 		})
+
+		t1.Run("should default to port 7687 neo4j://localhost", func(t *testing.T) {
+			driver, err := NewDriver("neo4j://localhost?x=y&a=b", NoAuth())
+			driverTarget := driver.Target()
+
+			assertNoError(t, err)
+
+			It("target port should be 7687", func() {
```
2hdddg: "It" is part of external test frameworks (ginkgo/gomega) that I plan to get rid of. 
assertStringEqual(t, driverTarget.Port(), "7687") should do the trick and only use standard Go testing

2hdddg (changes requested):
Nice! Mergable when comment fixed!

2hdddg (approved):

PR #125, opened 2020-07-28
Support 32-bit archs like ARM7

dhui:
Before this change, builds of [v1.8.0](https://github.com/neo4j/neo4j-go-driver/releases/tag/v1.8.0) failed:
```shell
$ GOARCH=arm GOARM=7 go build ./...
# github.com/neo4j/neo4j-go-driver/neo4j/internal/packstream
neo4j/internal/packstream/packer.go:130:10: constant 4294967295 overflows int
neo4j/internal/packstream/packer.go:167:9: constant 4294967296 overflows int
```
This PR fixes that

### DIFF
--- a/neo4j/internal/packstream/packer.go
+++ b/neo4j/internal/packstream/packer.go
@@ -127,7 +127,7 @@ func (p *Packer) writeListHeader(l int, shortOffset, longOffset byte) error {
 			hdr = hdr[:1+2]
 			hdr[0] = longOffset + 1
 			binary.BigEndian.PutUint16(hdr[1:], uint16(l))
-		case l < math.MaxUint32:
+		case uint32(l) < uint32(math.MaxUint32):
 			hdr = hdr[:1+4]
 			hdr[0] = longOffset + 2
 			binary.BigEndian.PutUint32(hdr[1:], uint32(l))
@@ -164,7 +164,7 @@ func (p *Packer) writeBytes(b []byte) error {
 		hdr = hdr[:1+2]
 		hdr[0] = 0xcd
 		binary.BigEndian.PutUint16(hdr[1:], uint16(l))
-	case l < 0x100000000:
+	case uint32(l) <= uint32(0xffffffff):
 		hdr = hdr[:1+4]
 		hdr[0] = 0xce
 		binary.BigEndian.PutUint32(hdr[1:], uint32(l))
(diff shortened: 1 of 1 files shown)


neo4j/internal/packstream/packer.go
```diff
@@ -127,7 +127,7 @@ func (p *Packer) writeListHeader(l int, shortOffset, longOffset byte) error {
 			hdr = hdr[:1+2]
 			hdr[0] = longOffset + 1
 			binary.BigEndian.PutUint16(hdr[1:], uint16(l))
-		case l < math.MaxUint32:
+		case uint32(l) < uint32(math.MaxUint32):
```
2hdddg: uint32(l) will truncate l to fit into uint32 and the comparison will always succeed
l needs to be int64 instead of int to compile and work properly on 32/64 bits systems.

neo4j/internal/packstream/packer.go
```diff
@@ -164,7 +164,7 @@ func (p *Packer) writeBytes(b []byte) error {
 		hdr = hdr[:1+2]
 		hdr[0] = 0xcd
 		binary.BigEndian.PutUint16(hdr[1:], uint16(l))
-	case l < 0x100000000:
+	case uint32(l) <= uint32(0xffffffff):
```
2hdddg: See above

2hdddg (changes requested):
Thanks for your pull request and good that you found this! We will look into compiling for 32 bits in our build pipelines to find these kind of errors early on!
You need to sign our CLA before we can merge your changes.

neo4j/internal/packstream/packer.go
dhui: Thanks for the feedback! Didn't know that `int64` [would work](https://groups.google.com/forum/#!topic/golang-nuts/mtnn-01Dh_I) on a 32-bit system

dhui:
> We will look into compiling for 32 bits in our build pipelines to find these kind of errors early on!

This would be great to prevent regressions!

> You need to sign our CLA before we can merge your changes.

I'd rather not sign the CLA and don't need the attribution but getting this issue fixed would be nice. This code change is trivial so you're welcome to take it and make it your own.

2hdddg:
Fixed in another PR

PR #129, opened 2020-08-03
Fix compilation error on 32 bits system

2hdddg:
Explicit cast to int64 to allow range check on 64 bits systems.


dhui:
Thanks for the quick fix!

PR #135, opened 2020-08-18
native go tests

martin-neotech:
auth_tokens_test.go
driver_test.go

### DIFF
--- a/neo4j/driver_test.go
+++ b/neo4j/driver_test.go
@@ -93,7 +93,6 @@ var uriSchemeTests = []struct {
 }{
 	{"bolt://", "bolt", "bolt://localhost:7687", false},
 	{"neo4j://", "neo4j", "neo4j://localhost:7687", true},
-	//{"bolt+routing://", "bolt+routing", "bolt+routing://localhost:7687", true},
 }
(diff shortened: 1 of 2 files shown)


neo4j/driver_test.go
```diff
@@ -84,158 +80,210 @@ func assertStringContain(t *testing.T, str, substr string) {
...
+	err = driver.Close()
+	if err != nil {
+		t.Errorf("should allow the close call on a closed driver")
+	}
+
+}
+
+func TestDriverSessionCreation(t *testing.T) {
+
+	t.Run("case one", func(t1 *testing.T) {
```
2hdddg: These looks like they should be table driven. Look here for an example: https://github.com/golang/go/wiki/TableDrivenTests
There are also examples within this codebase (probably within internal/...).
The auth stuff could also easily be written with thie approach, just comparing maps instead of comparing each entry in the maps

2hdddg (changes requested):

neo4j/driver_test.go
martin-neotech: updated

PR #138, opened 2020-08-27
Use system database when retrieving routing table

### LINKED ISSUE #137: Unable to retrieve routing table for user 'x' when user 'x' have no permission to default database
In setups where the user used when creating the driver doesn't have access to the system default database, the routing fails:
Unable to retrieve routing table from dbserver:7687: Server error: [Neo.ClientError.Security.Forbidden] Database access is not allowed for user 'x' with roles []."

2hdddg:
fix #137

### DIFF
--- a/neo4j/internal/bolt/bolt4.go
+++ b/neo4j/internal/bolt/bolt4.go
@@ -618,13 +618,19 @@ func (b *bolt4) GetRoutingTable(database string, context map[string]string) (*db
 		return nil, err
 	}
 
+	// The query should run in system database, preserve current setting and restore it when
+	// done.
+	originalDatabaseName := b.databaseName
+	b.databaseName = "system"
+	defer func() { b.databaseName = originalDatabaseName }()
+
+	// Query for the users default database or a specific database
 	const (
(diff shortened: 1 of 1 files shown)


neo4j/internal/bolt/bolt4.go
```diff
@@ -618,13 +618,19 @@ func (b *bolt4) GetRoutingTable(database string, context map[string]string) (*db
 		return nil, err
 	}
 
+	// The query should run in system database, preserve current setting and restore it when
+	// done.
+	originalDatabaseName := b.databaseName
+	b.databaseName = "system"
```
unterstein, line 624: Shouldn't this be the passed in variable `database`?

The routing table will be different for each database, so I don't think just `system` is the right call here. I guess we need to something like:

`b.databaseName = database` with the added deferred func to restore the previous one.

2hdddg: b.databaseName is the database that is used to execute within, there is a parameter to getRoutingTable that is set to database, that is we query for routing for "database" in "system"
It would be great if you could verify it with your setup. I'm working on another patch to include in the release.

Issue #142, opened 2020-08-30
[Feature Request] : (Enhancement): NeoError 


grizzly-monkey:
Currently there generic `error` interface that is used to throw error.
As well as the result errors are not part of the error but need to explicitly check record.
`record.Err()`

Instead of this can we have a checked `NeoError` which does proper error mapping with error `code` , `message` etc...
also moving the error from record interface and `Session.Run` itself maps internally to avoid checking the `error`.

**Neo4j Version:** 4.x Community  
**Neo4j Mode**: Single instance  
**Driver version**: Go driver 1.8.2  
**Operating System:** Ubuntu 16.10 on AWS

2hdddg:
Hi @grizzly-monkey 
I'm not sure exactly what you mean with some of these suggestions.

I cannot see us changing the driver methods to not return the "generic error" interface, I haven't seen any Go package that doesn't use standard error handling. But that is maybe not what you suggest. The 1.8 driver is backwards compatible with 1.7 and shouldn't change much, there is ongoing work on a 4.0 driver with quite a lot of API changes, in this version the database error will be made public so you could check if err is of that type. Please look at 4.0-dev branch for more info..

Not sure what you mean with check record error, there is no record.Err function. Do you mean that you need to check result.Err() before retrieving a record? But that isn't the case, if result.Next returns true, there is a record, you only need to check result.Err when result.Next returns false.

On 1.7, database errors were not returned until starting iterating, so on 1.8 session.Run sometimes "delays" the error to be backwards compatible. On upcoming 4.0 driver this will not be the case. Not sure if this is what you mean with "Session.Run itself maps internally..".

grizzly-monkey:
https://github.com/mongodb/mongo-go-driver/blob/master/mongo/errors.go
Was talking about these kinda error wrapper instead of return `errors.New("some error")`

```
type CommandError struct {
	Code    int32
	Message string
	Labels  []string // Categories to which the error belongs
	Name    string   // A human-readable name corresponding to the error code
	Wrapped error    // The underlying error, if one exists.
}

type WriteConcernError struct {
	Name    string
	Code    int
	Message string
	Details bson.Raw
}
etc...
```


for record.Err()

```
res, neoErr := dao.client.Session().Run(dao.cypher.TreeQuery(), map[string]interface{}{
		"NUMBER": partNumber,
		"REVISION":    revision,
	})

	if neoErr != nil { 
		return nil, mappers.MapNeoErr(neoErr)
[...]

PR #143, opened 2020-09-01
Allow access to DatabaseError code

marvinkite:
When I try to implement the error handling I find it hard to know what kind of error the Driver returns. As  #142 also suggested some improvement I think this simply allows to know what kind of error my code should deal with. 

For instance the code could be `Neo.ClientError.Schema.ConstraintValidationFailed` but through the `error` interface I can access only the string message `Server error: [Neo.ClientError.Schema.ConstraintValidationFailed] Node(6) already exists with label `. I think accessing the Code directly rather than parsing the string improves the usability of the library. 

The error types are internal which makes the error handling impossible so I agree with the need to improve it some way. These utility methods could be one way I agree some types of error messages could be internal because the Client application don't need to be concerned but when it's related to the consequences of Cypher and the Parameters so the client app shell handle it _(Like Constraint or Security)_ the error could expose more than just a string message.

### DIFF
--- a/neo4j/error.go
+++ b/neo4j/error.go
@@ -94,3 +94,13 @@ func IsServiceUnavailable(err error) bool {
 	_, is = err.(*pool.PoolFull)
 	return is
 }
+
+// GetDatabaseErrorCode is a utility method to check if the provided error is DatabaseError and if so
+// it gets the error code.
+func GetDatabaseErrorCode(err error) string {
+	dbErr, is := err.(*db.DatabaseError)
+	if !is {
+		return ""
+	}
+	return dbErr.Code
+}


2hdddg:
Hi @marvinkite 
Thanks for your pull request, you need to sign the CLA for the PR to get merged.
The database error struct is planned to be moved from internal package to public in 4.0, future for these error check functions are not determined yet. The approach you take here on 1.8 looks nice but is probably not the way on 4.0 (which means another deprecated API).

neo4j/error.go
```diff
@@ -94,3 +94,13 @@ func IsServiceUnavailable(err error) bool {
...
+
+// GetDatabaseErrorCode is a utility method to check if the provided error is DatabaseError and if so
+// it gets the error code.
+func GetDatabaseErrorCode(err error) string {
+	dbErr, is := err.(*db.DatabaseError)
+	if !is {
+		return ""
+	}
+	return dbErr.Code
+}
```
grizzly-monkey: add eol.

marvinkite: Fixed it and signed the CLA but I have no idea why the check fails.

2hdddg: Still CLA problems

marvinkite: Is there anything I can do to resolve the CLA problem? I followed the process to send the email. I cause could be the first commit does not contains the name or email but the second has the correct github username so I don't know how the check performed.

2hdddg: The build needs to rerun to check again and it triggers on commits.

We are very thankful for pull request and this one is nicely done but 1.8 is not the focus right now. Can you please have a look on 4.0-dev branch and see if the work done there solves your problem. 4.0 is on it's way, hopefully alpha pretty soon.

2hdddg:
If you really need this functionality on 1.8 feel free to open again.

PR #167, opened 2020-11-08
(WIP) Minimal invasive pre- and post-hooks for instrumentation and tracing

noctarius:
This Pull Request is a first draft for discussion on pre and post hooks around the actual execution of a query. Such functionality is used for distributed tracing tools and additional logging.

Any discussion around that topic is appreciated.

### DIFF
--- a/neo4j/hooks.go
+++ b/neo4j/hooks.go
@@ -1,0 +1,7 @@
+package neo4j
+
+import "net/url"
+
+type PreQueryHook = func(cypher string, params map[string]interface{}, target url.URL, config *TransactionConfig)
+
+type PostQueryHook = func(err error, target url.URL, config *TransactionConfig)
--- a/neo4j/transaction_config.go
+++ b/neo4j/transaction_config.go
@@ -19,7 +19,10 @@
 
 package neo4j
 
-import "time"
+import (
+	"context"
+	"time"
+)
 
 // TransactionConfig holds the settings for explicit and auto-commit transactions. Actual configuration is expected
 // to be done using configuration functions that are predefined, i.e. 'WithTxTimeout' and 'WithTxMetadata', or one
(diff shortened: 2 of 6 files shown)


neo4j/transaction_config.go
```diff
@@ -29,6 +32,8 @@ type TransactionConfig struct {
 	Timeout time.Duration
 	// Metadata is the configured transaction metadata that will be attached to the underlying transaction.
 	Metadata map[string]interface{}
+	// Context is the context to govern the execution
+	Context context.Context
```
2hdddg, line 36: Support for context is a different story and needs to go a lot deeper. I agree that it is needed but it is misplaced in this context (pun intended).

neo4j/hooks.go
```diff
@@ -0,0 +1,7 @@
+package neo4j
+
+import "net/url"
+
+type PreQueryHook = func(cypher string, params map[string]interface{}, target url.URL, config *TransactionConfig)
```
2hdddg, line 5: Does it make sense to implement just one or the other? If not an interface should be used.
Passing config as a pointer allows for modifications which would be another way of modifiying the tx, intended or unintended? Why would instrumentation hooks need to modify the config?
No need for aliasing in this case.

noctarius: Not perfect, but intended, since I'd have to change the context (actually overriding it with a new instance, which now contains the current span for the execution (to retrieve it in the second hook and finish it)

neo4j/transaction_config.go
noctarius: It is the least invasive version, not breaking the API. Happy for any better way to get it down there :) A different way may solve the second of your questions, too :)

2hdddg: Ok, since I didn't see the consuming side of the instrumentation it wasn't obvious that you used context to store data between the before/after hooks.
I really don't want to add Context just for implementing a placeholder for instrumentation state, it will look like the driver supports contexts in general which it currently doesn't.
Makes more sense to support context first and then add any instrumenation on top of that if context should be part of the instrumentation api.

2hdddg:
Hi @noctarius 
Thanks very much for all valuable inputs on instrumentation! We think we need to do this in steps, first support context in driver operations (maybe add separate interfaces on API level for that...). When that is in place we can add instrumentation hooks that takes the context as a parameter.

PR #184, opened 2021-01-26
Fix ownership issue of byte slice used as a buffer

### LINKED ISSUE #183: Query Not Returning Expected Results and Returning Zeros
**Neo4j Version:** 4.2.1  
**Neo4j Mode**: Using Aura on GCP
**Driver version**: Go driver 4.2.1  
**Operating System:** Using Aura on GCP

### Steps to reproduce
I apologize since we can not reproduce this outside of our production DB, which contains sensitive data. We are running a larger query where about half way through the return data, it starts to return all Zeros for the return fields. EDIT: We have found that one of the results is chunked, and that is the result that it is getting corrupted.
### Expected behavior
The query should return valid data. When connecting to our DB from the Node driver, and running the same query we get expected results.
### Actual behavior
The query returns valid data to a certain point, then begins to return all zeros. We are not seeing anything in the logs.

voutilad:
This change should fix #183. It appears that when we read a bolt message into the buffer, accounting for dechunking, a reference to that byte slice is then passed to the hydrator. However, since it's a reference to the same underlying buffer using by the "incoming" struct for processing bolt messages on the stream, it's possible data corruption can occur.

This might not be the best approach as it won't preserve any highwater-mark behavior that the driver currently has (because the dechunker reallocates the byte slice if the message size grows large enough). Maybe it's worth allocating a slice that's the same size as the old buffer to prevent future allocations?

> Note: it's not possible to just zero the existing `i.buf` because it's a shared reference with the `buf` passed to `i.hyd.hydrate()`

### DIFF
--- a/neo4j/internal/bolt/incoming.go
+++ b/neo4j/internal/bolt/incoming.go
@@ -33,8 +33,9 @@ func (i *incoming) next(rd io.Reader) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Reuse buffer for next dechunk
-	i.buf = buf
+	// Create a new buffer for the incoming stream so the hydrator can own the
+	// current byte slice.
+	i.buf = make([]byte, 4096)
 
 	return i.hyd.hydrate(buf)
 }


2hdddg:
@voutilad I fixed the actual problem related to how the reused buffer was resized during dechunking and added some unit tests that found the bad behavior. Also tried the example query in program before and after the fix. 
Thanks for a perfect description of the error and way to reproduce it!

PR #188, opened 2021-02-09
Use transaction functions in README example

fbiville:


fbiville:
The sample becomes significantly larger, but I think it improves a few things:

1. the code is directly reusable
1. imports are shown (including the esp. important driver v4 package)
1. it uses transaction functions instead of `Session#Run`
1. error handling around closing is slightly improved as well (as Close can itself error)

Regarding point 1, the only reason I did not write a `main` function is to avoid using `panic` (which imo is usually only valid in `main`, but not so much elsewhere and I don't want people to copy that)

If you think it's too verbose, I'm happy to reduce bits you think are not necessary.

README.md
```diff
@@ -29,36 +29,73 @@ Connect, execute a statement and handle results
...
+    // Neo4j 4.0, defaults to no TLS therefore use bolt:// or neo4j://
+    // Neo4j 3.5, defaults to self-signed cetificates, TLS on, therefore use bolt+ssc:// or neo4j+ssc://
+    dbUri := "neo4j://localhost:7687"
+    driver, err := neo4j.NewDriver(dbUri, neo4j.BasicAuth("username", "password", ""))
+    if err != nil {
+        return err
+    }
+    // Handle driver lifetime based on your application lifetime requirements  driver's lifetime is usually
+    // bound by the application lifetime, which usually implies one driver instance per application
+    defer unsafeClose(driver)
```
2hdddg: Nitpick but I wouldn't care about handling this error, simpler to just do a defer.

README.md
```diff
@@ -29,36 +29,73 @@ Connect, execute a statement and handle results
...
+    
+    // Sessions are short-lived, cheap to create and NOT thread safe. Typically create one or more sessions
+    // per request in your web application. Make sure to call Close on the session when done.
+    // For multi-database support, set sessionConfig.DatabaseName to requested database
+    // Session config will default to write mode, if only reads are to be used configure session for
+    // read mode.
+    session := driver.NewSession(neo4j.SessionConfig{})
+    defer unsafeClose(session)
```
2hdddg: Same as above, whatever error you get here shouldn't matter..

README.md
```diff
@@ -29,36 +29,73 @@ Connect, execute a statement and handle results
...
+    // Handle driver lifetime based on your application lifetime requirements  driver's lifetime is usually
+    // bound by the application lifetime, which usually implies one driver instance per application
+    defer unsafeClose(driver)
+    
+    // Sessions are short-lived, cheap to create and NOT thread safe. Typically create one or more sessions
+    // per request in your web application. Make sure to call Close on the session when done.
+    // For multi-database support, set sessionConfig.DatabaseName to requested database
+    // Session config will default to write mode, if only reads are to be used configure session for
+    // read mode.
+    session := driver.NewSession(neo4j.SessionConfig{})
```
2hdddg, line 65: Maybe make the multi-database support even more visible by setting a value here... Not sure

README.md
```diff
@@ -29,36 +29,73 @@ Connect, execute a statement and handle results
...
+        return err
+    }
+    fmt.Printf("%v\n", result)
+    return nil
+}
 
-// Handle driver lifetime based on your application lifetime requirements  driver's lifetime is usually
-// bound by the application lifetime, which usually implies one driver instance per application
-defer driver.Close()
+func createItem(tx neo4j.Transaction) (interface{}, error) {
```
2hdddg: Maybe state the importance of returning the drivers native errors here when errors originate from driver calls (no wrapping) otherwise the retry mechanism won't work properly. (this should be redesigned, the driver knows the error that happened...)

README.md
```diff
@@ -29,36 +29,73 @@ Connect, execute a statement and handle results
...
+    // Session config will default to write mode, if only reads are to be used configure session for
+    // read mode.
+    session := driver.NewSession(neo4j.SessionConfig{})
+    defer unsafeClose(session)
+    
+    result, err := session.WriteTransaction(createItem)
+    if err != nil {
+        return err
+    }
+    fmt.Printf("%v\n", result)
```
2hdddg: Show the ugly casting needed?

2hdddg (approved):
Excellent work! I have some small comments but they are all optional and maybe not even good ideas.

README.md
fbiville: there are already some casts in `createItem`. Shall we show more of them? ^^

2hdddg: Nah, maybe overkill... Meant the casting of result to *Item to make it a bit more realistic

README.md
fbiville: I can actually improve the error handling by throwing `%w` into the mix and using a named `err` return.
But are you saying driver close errors should generally be ignored?

README.md
fbiville: That part was left unchanged. Let's keep it like this for now?

README.md
2hdddg: I don't think they should have returned an error in the first place. Driver.Close never returns an error and Session.Close only returns an error if there was an error opening the session (but that would be visible to any other operation before Close) or if user is closing the session with a pending transaction.

fbiville:
I went ahead and introduce a main function.
That way, code is actually reusable and the driver is instantiated there, enforcing the idea of its "singleton-ness".

2hdddg (approved):
Great improvement!

Issue #190, opened 2021-02-15
Potential Deadlock in Router in clustered environment


marvinkite:
Potential deadlock condition in router:

[https://github.com/neo4j/neo4j-go-driver/blob/4.2/neo4j/internal/router/router.go#L78](https://github.com/neo4j/neo4j-go-driver/blob/4.2/neo4j/internal/router/router.go#L78)

```golang
func (r *Router) getTable(database string) (*db.RoutingTable, error) {
	now := r.now()

	r.dbRoutersMut.Lock() // All my request hangs here now.
	defer r.dbRoutersMut.Unlock()

	dbRouter := r.dbRouters[database]
	if dbRouter != nil && now.Unix() < dbRouter.dueUnix {
		return dbRouter.table, nil
	}

	var (
		table *db.RoutingTable
		err   error
	)

	// Try last known set of routers if there are any
	if dbRouter != nil && len(dbRouter.table.Routers) > 0 {
		routers := dbRouter.table.Routers
		r.log.Infof(log.Router, r.logId, "Reading routing table for '%s' from previously known routers: %v", database, routers)
		table, err = readTable(context.Background(), r.pool, database, routers, r.routerContext)
	}
```

**Neo4j Version:** Aura Cloud  
**Neo4j Mode**: Cluster  
**Driver version**: Go driver 4.2.2  
**Operating System:** Linux,OSX  

With a single instance there is no problem and the service scales well, when connecting to Aura and read mode routes between two nodes after a minute under performance test we hit this problem and we must restart the server. 

Would be nice to add the context to each call so the caller could set the deadline and if any subsequent call times out it could release the lock. 

Instead of this:

```golang
table, err = readTable(context.Background(), r.pool, database, routers, r.routerContext)
```

What we see now all request is queuing up and there is no way yo cancel them and free up those queries which are blocked at the lock but the client already abandoned the response and timed out. 

I could contribute and help to test and fix because this is crucial for us but with this deadlock we can't use this driver.

2hdddg:
I have a look tomorrow at this, it would be really great to have logs for when this happens.

marvinkite:
I enabled the logging and  with shorter timeout and also with the context timeout the server survived the deadlock and there were many connection timeout to the Aura server. I will save the logs and sent it.

```golang
driver, err = neo4j.NewDriver(c.neo4jHost, c.neo4jToken,
	CustomLogger(c.log), func(config *neo4j.Config) {
[...]

PR #191, opened 2021-02-15
Add context.Context to function arguments

marvinkite:
Adding `context.Context` to each function. This is a proposal to follow a practice  to use context as function parameter. 

I know this breaks the API but I don't know how important to keep the API unchanged.

**Potential pros:**

* Enable tracing of driver. In our project we use OpenTelemetry and this would greatly simplify the integration.
* Allow to control the operation timeout or cancel.

If the HTTP server has a request timeout and under heavy load the request all waiting for the pool to return a Connection but it takes more time than the timeout, there is no way not to cancel the operation. This could allow but there are still a lot to do until. 

I intend to solve the #190 problem that makes the system non response under high load.

### DIFF
--- a/neo4j/directrouter.go
+++ b/neo4j/directrouter.go
@@ -19,20 +19,22 @@
 
 package neo4j
 
+import "context"
+
 // A router implementation that never routes
 type directRouter struct {
 	address string
 }
 
-func (r *directRouter) Readers(database string) ([]string, error) {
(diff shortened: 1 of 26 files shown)


2hdddg:
Support for context should be added at some point but breaking the API is unfortunately not an option. Besides the solution to the problem in #190  is not context:s..

PR #194, opened 2021-02-17
Fixes problem with discarding big streams

2hdddg:


fbiville (approved):
It might be worth adding a more granular test around `discardStream`, to make sure the `SUCCESS` response of the `DISCARD` request is acknowledged.

Happy to try adding that if you think it makes sense.

neo4j/internal/bolt/stream.go
```diff
@@ -136,10 +132,17 @@ func (o *openstreams) resume(s *stream) {
 	o.curr = s
 }
 
-// Internal, "removes" the stream by setting it's corresponding entry to nil.
+// "removes" the stream by disabling it's key and removing it from the count of streams
```
fbiville: nit: s/it's/its/

PR #198, opened 2021-03-08
Refactored stream counter

2hdddg:
Simpler implementation, does not mess with callbacks and tries to be
smart with allocations.

### DIFF
--- a/neo4j/internal/bolt/bolt4.go
+++ b/neo4j/internal/bolt/bolt4.go
@@ -109,17 +109,12 @@ func NewBolt4(serverName string, conn net.Conn, log log.Logger) *bolt4 {
 		packer:  &packstream.Packer{},
 		onErr:   func(err error) { b.setError(err, true) },
 	}
-	// Setup open streams. Errors reported to callback are assertion like errors, let them
-	// bubble up to kill them off when they happen, alternatively they could be logged.
-	b.streams.onAssertFail = func(err error) {
-		b.setError(err, false)
-	}
-	b.streams.onClose = func(s *stream) { // Current stream succesfully closed (no error)
-		if len(s.sum.Bookmark) > 0 {
-			b.bookmark = s.sum.Bookmark
-		}
-	}
-	b.streams.onEmpty = func() { // All streams closed (succesfully or with error)
+
+	return b
+}
+
+func (b *bolt4) checkStreams() {
+	if b.streams.num <= 0 {
 		// Perform state transition from streaming, if in that state otherwise keep the current
 		// state as we are in some kind of bad shape
 		switch b.state {
@@ -129,8 +124,6 @@ func NewBolt4(serverName string, conn net.Conn, log log.Logger) *bolt4 {
 			b.state = bolt4_ready
 		}
 	}
-
-	return b
 }
 
 func (b *bolt4) ServerName() string {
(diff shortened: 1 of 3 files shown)


neo4j/internal/bolt/bolt4.go
```diff
@@ -109,17 +109,12 @@ func NewBolt4(serverName string, conn net.Conn, log log.Logger) *bolt4 {
...
-			b.bookmark = s.sum.Bookmark
-		}
-	}
-	b.streams.onEmpty = func() { // All streams closed (succesfully or with error)
+
+	return b
+}
+
+func (b *bolt4) checkStreams() {
+	if b.streams.num <= 0 {
```
fbiville, line 117: can it be strictly lower than 0?

neo4j/internal/bolt/bolt4.go
```diff
@@ -766,8 +763,12 @@ func (b *bolt4) receiveNext() (*db.Record, bool, *db.Summary) {
 		sum.ServerVersion = b.serverVersion
 		sum.ServerName = b.serverName
 		sum.TFirst = b.tfirst
+		if len(sum.Bookmark) > 0 {
```
fbiville, line 766: the previous version used to do this when a stream was closed (via `b.streams.onClose`)
At first glance, this seems to change the bookmark behavior.
It does not, because closing a session or switching to another stream forces the current stream consumption leading to this "success" code path here?

neo4j/internal/bolt/stream.go
```diff
@@ -100,20 +96,11 @@ func (o *openstreams) attach(s *stream) {
...
-	}
-
-	if sum != nil {
-		o.curr.sum = sum
-		o.onClose(o.curr)
-	} else if err != nil {
-		o.curr.err = err
-	} else {
-		o.onAssertFail(assertDetachUncomplete)
 		return
```
fbiville, line 99: should there an error message in that case?

fbiville (commented):
first pass: newbie questions :)
TODO: review test changes

neo4j/internal/bolt/bolt4.go
2hdddg: Not really... it Is an error that might be worth logging as debug

neo4j/internal/bolt/bolt4.go
2hdddg: There will only be a bookmark when hasMore is false so it shouldn't be an issue. Might be that we don't even need an if len(...) anymore. It was needed before because it was invoked in more situations. Good catch though!

neo4j/internal/bolt/stream.go
2hdddg: The hook could log.
I removed the error hook and the error reporting in another PR, it is programming error when those happens and should be detected by unit tests on another level.

neo4j/internal/bolt/stream_test.go
```diff
@@ -92,86 +92,52 @@ func TestStream(ot *testing.T) {
...
 		AssertNotNil(t, streams.curr)
-
-		// Detaching uncompleted stream should fail
-		streams.detach(nil, nil)
-		AssertFalse(t, emptied)
-		AssertFalse(t, closed)
-		AssertTrue(t, failed)
-		AssertNotNil(t, streams.curr)
-		reset()
+		AssertIntEqual(t, streams.num, 1)
```
fbiville, line 105: nit: check if `streams.curr.err` is nil

fbiville (approved):

PR #209, opened 2021-04-09
Send ROUTE with bookmarks

fbiville:


2hdddg (approved):
Looks fine, if the bookmarks are needed for procedure call getRoutingTable then you need to add it in bolt4.callgetRoutingTable and bolt3 as well.

fbiville:
`bolt3` is not concerned by this change, since bookmarks are only needed in the scenario of a previous command creating a database (multi-database is not available in Neo4j 3.x-).
I added a change to include bookmarks when calling the procedure. Thanks!

2hdddg (approved):

PR #210, opened 2021-04-14
Treat expired auth errors as fatal 

fbiville:


2hdddg (changes requested):
Since every connection has a timestamp from when it was used you can use that in combination with a timestamp in the pool to know which connections to remove. You shouldn't need to tag every connection that way.

neo4j/internal/pool/pool.go
```diff
@@ -414,3 +418,21 @@ func (p *Pool) Return(c db.Connection) {
 		p.log.Warnf(log.Pool, p.logId, "Server %s not found", serverName)
 	}
 }
+
```
2hdddg: Both of these are accessing structures written to from other threads, will not work.

fbiville:
Changed the PR completely now @2hdddg, let me know if that approach is better.
I'll add more test coverage as soon as you think the approach is OK.

2hdddg (approved):
That should do it!
Perhaps a test on bolt4 to ensure it. That should be enough since there is no extra logic needed in the pool

2hdddg (approved):

PR #213, opened 2021-05-04
Fix tests further to ba062545b93fbaad2f84787497e00bfd1f23e3ba changes

fbiville:


neo4j/test-integration/dbconn_test.go
```diff
@@ -65,7 +65,7 @@ func makeRawConnection(logger log.Logger) (dbserver.DbServer, db.Connection) {
 }
 
 func BenchmarkQuery(b *testing.B) {
-	_, conn := makeRawConnection(&log.Console{Debugs: true, Errors: true, Infos: true, Warns: true})
+	_, conn := makeRawConnection(&log.Console{Debugs: false, Errors: true, Infos: true, Warns: true})
```
fbiville, line 68: either that or I introduce some kind of log id-based filtering into the logger config. WDYT @2hdddg?

2hdddg: Not sure what the question is. The benchmarks doesn't run anywhere as far as I know

fbiville: it runs in CI and with the new messages being logged in DEBUG, the tests would time out.

fbiville: (notably the `RECORD` message)

2hdddg: Are you sure you run the benchmarks in CI? You shouldn't, it might be that they are but that is a mistake

fbiville: In any case, they revealed a real issue: when pulling many records, (debug) logging can become a bottleneck. What do I do about this (since it used to be not so severe before)?

fbiville: As discussed offline, merging this for now and creating another PR to move the Bolt logging to a separate utility

PR #214, opened 2021-05-06
Server info

fbiville:


neo4j/resultsummary.go
```diff
@@ -179,6 +182,14 @@ type resultSummary struct {
 	params map[string]interface{}
 }
 
+func (s *resultSummary) Agent() string {
+	return s.sum.Agent
+}
+
+func (s *resultSummary) ProtocolVersion() string {
```
2hdddg: Do we need to return a string? Could be 
ProtocolVersion() (major int, minor int)
Which would be cheaper internally and reduce need for parsing on consuming end

2hdddg (commented):
Minor comment...

2hdddg (commented):
Yes that will work.
One picky thing is that there are no exports from db package in the neo4j api to this point. This would make users have to import neo4j/db.
Cleanest would be to maybe keep minor/minor in Summary and have ProtocolVersion struct (no need for unnecessary interface as all the other ones...) and convert in ResultSummary.ProtocolVersion

2hdddg (approved):

PR #220, opened 2021-05-17
testkit-backend: treat 'Invalid transaction handle' as driver error

bigmontz:
This kind of errors were being wrongly considered backend errors, It could make difficult to assert situations of wrong driver usage like commit in a already commited transaction and checks if the driver thorws an error.

### DIFF
--- a/testkit-backend/backend.go
+++ b/testkit-backend/backend.go
@@ -112,7 +112,8 @@ func (b *backend) writeError(err error) {
 	isDriverError := neo4j.IsNeo4jError(err) ||
 		neo4j.IsUsageError(err) ||
 		neo4j.IsConnectivityError(err) ||
-		neo4j.IsTransactionExecutionLimit(err)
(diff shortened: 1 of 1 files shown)


testkit-backend/backend.go
```diff
@@ -112,7 +112,8 @@ func (b *backend) writeError(err error) {
 	isDriverError := neo4j.IsNeo4jError(err) ||
 		neo4j.IsUsageError(err) ||
 		neo4j.IsConnectivityError(err) ||
-		neo4j.IsTransactionExecutionLimit(err)
+		neo4j.IsTransactionExecutionLimit(err) ||
+		err.Error() == "Invalid transaction handle"
```
fbiville, line 116: It seems to me this error should be created as a usage error, not an ad-hoc one as it is currently the case.
Given the current implementation, this seems like the way to go.

fbiville (approved):

PR #223, opened 2021-05-20
Fixed path parse error

erictg:
Fixed error with path parsing where it would return a zero path if no relationships were found. An example query that could reproduce this is `match p=(n) return n` (assuming there is one node in the graph)

I also added test to verify this error is resolved.

A few other files were updated because I ran gofmt


.gitignore
```diff
@@ -6,6 +6,9 @@
 
 # GoLang
 .idea/
+# remove this if you want to use go modules at some point
```
fbiville: there is a `go.mod`/`go.sum` at the project root already

neo4j/internal/bolt/path.go
```diff
@@ -34,7 +34,13 @@ type relNode struct {
 func buildPath(nodes []dbtype.Node, relNodes []*relNode, indexes []int) dbtype.Path {
 	num := len(indexes) / 2
 	if num == 0 {
-		return dbtype.Path{}
+		var path dbtype.Path
+		// add nodes to the path if no relationships but there is a node
+		// if num is 0 there probably isn't rel nodes
+		if nodes != nil && len(nodes) != 0 {
```
fbiville, line 40:
```suggestion
		if len(nodes) != 0 {
```

neo4j/internal/bolt/path_test.go
```diff
@@ -57,6 +57,24 @@ func TestBuildPath(ot *testing.T) {
...
+			name: "One node",
+			path: dbtype.Path{
+				Nodes: []dbtype.Node{dbtype.Node{Id: 1}},
+			},
+			nodes:    []dbtype.Node{dbtype.Node{Id: 1}},
+			relNodes: []*relNode{},
+			indexes:  []int{},
+		},
+		{
+			name: "Many nodes, no relationships",
```
fbiville: This case normally can not happen

fbiville:
Now superseded by https://github.com/neo4j/neo4j-go-driver/pull/226

PR #249, opened 2021-07-28
Fix profile child computation

fbiville:
A range variable "lives" at the same virtual address during the
whole for scope. Without this fix, this means the plan children
slice will be made of n times the same profile plan instance (i.e.
the last one).


robsdedude (approved):
Looks GO to me.

bigmontz (approved):
😢

PR #259, opened 2021-08-25
Connection read timeout hint

fbiville:
Companion PR: https://github.com/neo4j-drivers/testkit/pull/233

### DIFF
--- a/neo4j/internal/bolt/bolt4.go
+++ b/neo4j/internal/bolt/bolt4.go
@@ -74,25 +74,26 @@ func (i *internalTx4) toMeta() map[string]interface{} {
 }
 
 type bolt4 struct {
-	state         int
-	txId          db.TxHandle
-	streams       openstreams
-	conn          net.Conn
-	serverName    string
-	out           outgoing
-	in            incoming
-	connId        string
-	logId         string
-	serverVersion string
-	tfirst        int64       // Time that server started streaming
-	pendingTx     internalTx4 // Stashed away when tx started explicitly
-	hasPendingTx  bool
-	bookmark      string // Last bookmark
-	birthDate     time.Time
-	log           log.Logger
-	databaseName  string
-	err           error // Last fatal error
-	minor         int
+	state                 int
+	txId                  db.TxHandle
+	streams               openstreams
+	conn                  net.Conn
+	serverName            string
+	out                   outgoing
+	in                    incoming
+	connId                string
+	logId                 string
+	serverVersion         string
+	tfirst                int64       // Time that server started streaming
+	pendingTx             internalTx4 // Stashed away when tx started explicitly
+	hasPendingTx          bool
+	bookmark              string // Last bookmark
+	birthDate             time.Time
+	log                   log.Logger
+	databaseName          string
+	err                   error // Last fatal error
+	minor                 int
+	connectionReadTimeout time.Duration
 }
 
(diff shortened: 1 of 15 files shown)


neo4j/internal/bolt/bolt4.go
```diff
@@ -182,8 +184,9 @@ func (b *bolt4) receiveMsg() interface{} {
 		return nil
 	}
 
-	msg, err := b.in.next(b.conn)
+	msg, err := b.in.next(b.conn, b.resetConnectionReadDeadline)
 	b.setError(err, true)
+	b.resetConnectionReadDeadline()
```
bigmontz: Seems like we are sharing the responsibility of calling `resetConnectionReadDeadline` between `Bolt` and `dechunker`.  This responsibility should be or pushed to the `dechunker` or keep it in bolt. I tend to prefer to push it to the lower level (`dechunker`) since this one is handling the actual reading.

fbiville: yep, good idea... this first draft felt messy.
The only drawback I see is that the dechunker will now have to accept a `net.Conn` instead of the currently more generic `io.Reader`

fbiville: I pushed a new version. Migrating tests from `io.Reader` to `net.Conn` was a bit of a PITA, but I finally managed to do it.

neo4j/internal/bolt/dechunker_test.go
```diff
@@ -69,19 +70,26 @@ func TestDechunker(t *testing.T) {
...
+		}()
+		buf, msgBuf, err = dechunkMessage(serv, buf, -1, nil, "")
 		AssertNoError(t, err)
 		AssertLen(t, msgBuf, int(msg.size))
 		// Check content of buffer
 		b = 0
 		for i := range msgBuf {
 			if msgBuf[i] != b {
 				t.Errorf("Wrong content in buffer at %d, %d vs %d (message %d)", i, msgBuf[i], b, msgi)
+				AssertNoError(t, serv.Close())
```
bigmontz, line 92: Is that a way to unit test the `dechunker` reset timeout logic?

neo4j/internal/bolt/hydrator.go
```diff
@@ -241,6 +242,9 @@ func (h *hydrator) success(n uint32) *success {
 			succ.notifications = parseNotifications(l)
 		case "rt":
 			succ.routingTable = h.routingTable()
+		case "hints":
```
bigmontz, line 245: Is there test for the hint set?

bigmontz (approved):
🧀

neo4j/internal/bolt/dechunker_test.go
fbiville: I only relied on the existing Testkit tests. Is it worth adding a test here as well in your opinion?

neo4j/internal/bolt/hydrator.go
fbiville: Via testkit, yes.

fbiville: And now a unit test as well

neo4j/internal/bolt/dechunker_test.go
fbiville: And now there is an integration test as well

bigmontz (approved):
📦

PR #260, opened 2021-09-02
Close connections on cluster retryable errors

eastlondoner:
so that a new TCP connection is used to fetch new routing table

When server side routing is used with a load balancer using the same TCP connection will get you to the same back end. Even if the intervening load balancer has detected a problem it will not necessarily kill open tcp connections

closing the tcp connection and getting a new one after a cluster-retriable error ensures that you get a connection to a server that any intervening components think is healthy

### DIFF
--- a/neo4j/internal/retry/state.go
+++ b/neo4j/internal/retry/state.go
@@ -109,6 +109,7 @@ func (s *State) OnFailure(conn db.Connection, err error, isCommitting bool) {
 	if dbErr, isDbErr := err.(*db.Neo4jError); isDbErr {
 		if dbErr.IsRetriableCluster() {
 			// Force routing tables to be updated before trying again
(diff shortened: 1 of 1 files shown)


fbiville:
I think your fix is spot on.
I just pushed https://github.com/neo4j-drivers/testkit/pull/248, so we may detect a similar issue with other drivers.

eastlondoner:
@fbiville FYI I leave this in your hands to merge / close / amend as you see fit.

I just created this PR as a way to communicate my idea about a fix. I'm not intending to do actually do anything else with this PR

fbiville:
No problem, thanks for the initial fix. I'll wait for the Testkit PR to be validated before going further.

robsdedude:
This issue will have to be resolved in the SSR implementation.

PR #265, opened 2021-09-22
Rename query in result summary to match other drivers

robsdedude:
Deprecate ResultSummary.Statement() -> .Query()
Rename Statement interface -> Query
Deprecate Query.Params() -> .Parameters()

### DIFF
--- a/neo4j/resultsummary.go
+++ b/neo4j/resultsummary.go
@@ -44,7 +44,7 @@ const (
 type ResultSummary interface {
 	// Server returns basic information about the server where the statement is carried out.
 	Server() ServerInfo
-	// Deprecated: Use Query instead
+	// Deprecated: since 4.4, will be removed in 5.0. Use Query instead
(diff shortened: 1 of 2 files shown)


neo4j/resultsummary.go
```diff
@@ -44,8 +44,10 @@ const (
 type ResultSummary interface {
 	// Server returns basic information about the server where the statement is carried out.
 	Server() ServerInfo
-	// Statement returns statement that has been executed.
+	// Deprecated: Use Query instead
```
fbiville: shall we say when it's gonna be removed? (5.0, right?)

fbiville (approved):

neo4j/resultsummary.go
robsdedude: Sounds like a good idea. And yes, 5.0 seems the right time to remove it.

fbiville (approved):

PR #266, opened 2021-09-24
Do not raise error when closing a session with a pending tx

fbiville:
This aligns with the behavior of the other official drivers.


robsdedude (approved):
:zap: :no_entry_sign:

PR #269, opened 2021-09-28
Add support for user impersonation

fbiville:
This allows to impersonate other users at the session level, via
SessionConfig#ImpersonatedUser.

### DIFF
--- a/neo4j/internal/bolt/bolt4.go
+++ b/neo4j/internal/bolt/bolt4.go
@@ -45,11 +45,12 @@ const (
 const bolt4_fetchsize = 1000
 
 type internalTx4 struct {
-	mode         db.AccessMode
-	bookmarks    []string
-	timeout      time.Duration
-	txMeta       map[string]interface{}
-	databaseName string
+	mode             db.AccessMode
+	bookmarks        []string
+	timeout          time.Duration
+	txMeta           map[string]interface{}
+	databaseName     string
+	impersonatedUser string
 }
 
 func (i *internalTx4) toMeta() map[string]interface{} {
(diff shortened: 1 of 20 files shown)


neo4j/internal/bolt/bolt4.go
```diff
@@ -874,13 +886,25 @@ func (b *bolt4) Reset() {
...
+	if b.minor >= 4 {
+		b.out.appendRouteWithDbContext(context, bookmarks, routeDatabaseContext(database, impersonatedUser))
+		b.out.send(b.conn)
+		succ := b.receiveSuccess()
+		if b.err != nil {
+			return nil, b.err
+		}
+		routingTable := succ.routingTable
+		b.databaseName = routingTable.Database
+		b.impersonatedUserHomeDb = routingTable.Database
```
robsdedude, line 904: The `db` field in the response will contain the database for which the RT has been queried. This is not necessarily the home database of the user. However, it will certainly be their home db, when no database has been specified.

neo4j/internal/bolt/bolt4.go
```diff
@@ -983,3 +1007,26 @@ func (b *bolt4) initializeReadTimeoutHint(hints map[string]interface{}) {
...
+func (b *bolt4) setImpersonationConfig(config db.TxConfig, tx *internalTx4) {
+	tx.impersonatedUser = config.ImpersonatedUser
+	// b.databaseName may have been overwritten by the session
+	// else it's the same as the impersonated user's home DB
+	// in the latter case, do not include the database since it might be outdated
+	// indeed, the home DB might have switched between the routing table retrieval
+	// and the execution of BEGIN, so let the server resolve the home DB again instead
+	if tx.databaseName == b.impersonatedUserHomeDb {
+		tx.databaseName = ""
+	}
```
robsdedude, line 1031: In both places, where this function is called, `tx.databaseName` has been set to `b.databaseName`. So this could as well be `b.databaseName == b.impersonatedUserHomeDb`. Which seems to be always true according to https://github.com/neo4j/neo4j-go-driver/pull/269/files#diff-123e0e61642d0d876f8df56b27c5f4a262baa6ef45ba4ef003c337fdb6f55aceR903-R904

fbiville: It's actually not always true, since `SelectDatabase` might be called by the database provided by SessionConfig (see https://github.com/neo4j/neo4j-go-driver/blob/5a14c7024ca3203d89d54ae34bbfbc2886249401/neo4j/session.go#L372-L379), hence the check

neo4j/internal/bolt/bolt4.go
fbiville: so something like:
```go
if database != db.DefaultDatabase {
    b.impersonatedUserHomeDb = routingTable.Database
}
```
?
the routing table DB is still the database to specify in subsequent operations, right?

fbiville:
Closing this for now, as we're discussing another approach for the drivers.

PR #273, opened 2021-10-08
Feature: user impersonation

2hdddg:
Requires Neo4j 4.4.


robsdedude:
Waiting for https://github.com/neo4j-drivers/testkit/pull/272 to be merged, then re-running the TestKit pipeline. Should pass. Then it's ready to be merged.

neo4j/internal/bolt/bolt3.go
```diff
@@ -469,6 +472,9 @@ func (b *bolt3) Run(runCommand db.Command, txConfig db.TxConfig) (db.StreamHandl
 	if err := b.assertState(bolt3_streaming, bolt3_ready); err != nil {
 		return nil, err
 	}
+	if err := b.checkImpersonation(txConfig.ImpersonatedUser); err != nil {
```
fbiville, line 475: interesting, I probably would have left this out entirely :)

robsdedude: I think there is a TestKit test for this case. We definitely agreed that the drivers should error if the user tries to impersonate and a Bolt 3 connection gets picked up from the pool.

PR #291, opened 2021-12-03
Add result error handling code example

### LINKED ISSUE #272: Neo.ClientError.Statement.SemanticError is not thrown to app
I have in logs  DEBUG  [bolt4 bolt-204@localhost:7687] Neo4jError: Neo.ClientError.Statement.SemanticError (Cannot merge the following node because of null property value for 'id': (:Node_label {id: null})),  But it does not throw an error in session.Run() or tx.Run()

fbiville:
Fixes #272

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,22 @@ There are a few points that need to be highlighted:
 * The driver is thread-safe, while the session or the transaction is not thread-safe.
 
 ## Parsing Result Values
+
 ### Record Stream
 A cypher execution result is comprised of a stream of records followed by a result summary.
-The records inside the result can be accessed via `Next()`/`Record()` functions defined on `Result`. It is important to check `Err()` after `Next()` returning `false` to find out whether it is end of result stream or an error that caused the end of result consumption.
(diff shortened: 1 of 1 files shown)


bigmontz (approved):
🔢

Kubera2017:
@fbiville sorry I've haven't answered quickly. One thing that I wanted to emphasize (and this issue ticket was about it) is only-writing query. When we do it, we usually don't need any result, just without errors (imagine some "cloud function" and stream)

When we faced with it our code was:
```
tx, err := session.BeginTransaction(neo4j.WithTxTimeout(25 * time.Second))
if err != nil {panic}
db_result, err := tx.Run(query, params)
if err != nil {panic}
// close and return
```
And this code didn't catch Neo.ClientError.Statement.SemanticError, but formally tx.Run was completed

After that we added additionally .Consume():
```
_, err = db_result.Consume()
if err != nil {panic}
```

fbiville:
Thanks, I'll update the example or add another one.

PR #293, opened 2021-12-06
Add context.Context to relevant API

fbiville:


neo4j/internal/connector/connector.go
```diff
@@ -86,7 +86,7 @@ func (c Connector) Connect(address string, boltLogger log.BoltLogger) (db.Connec
 	}
 	config := tls.Config{InsecureSkipVerify: c.SkipVerify, RootCAs: c.RootCAs, ServerName: serverName}
 	tlsconn := tls.Client(conn, &config)
-	err = tlsconn.Handshake()
+	err = tlsconn.HandshakeContext(ctx)
```
fbiville, line 89: this is where I'm not sure... should I use another context param for this specific TLS handshake or not?

fbiville (commented):
Another open question is: do I need to expose a `context.Context` on the low-level chunker and dechunker since they write/read data and could possibly take a long time?
There is no direct support for this in Go's I/O package but we can set read/write deadline on the TCP connection. Creating a bridge API with `context.Context` which under the hood translate read/write deadline should be straightforward.

testkit-backend/backend.go
```diff
@@ -608,7 +606,6 @@ func (b *backend) handleRequest(req map[string]interface{}) {
 				"Optimization:ConnectionReuse",
 				"Optimization:ImplicitDefaultArguments",
 				"Optimization:PullPipelining",
-				"Temporary:ConnectionAcquisitionTimeout",
```
fbiville, line 611: @robsdedude this PR unfortunately conflicts with timeout configuration on the testkit backend side
if this PR is merged, timeout/deadlines/... would be configured through each API via `context.Context` object, so it won't be possible to set a timeout before calling the related operation

neo4j/test-integration/examples_test.go
```diff
@@ -540,21 +518,11 @@ func createDriverWithCustomizedConnectionPool(uri, username, password string) (n
 	return neo4j.NewDriver(uri, neo4j.BasicAuth(username, password, ""), func(config *neo4j.Config) {
 		config.MaxConnectionLifetime = 30 * time.Minute
 		config.MaxConnectionPoolSize = 50
-		config.ConnectionAcquisitionTimeout = 2 * time.Minute
 	})
 }
 
 // end::config-connection-pool[]
 
-// tag::config-connection-timeout[]
```
fbiville, line 549: not sure if that's OK to just remove the entire tag

testkit-backend/backend.go
robsdedude: Couldn't you store the timeout that TestKit requires somewhere in you TestKit back end to then later pass it into a function call?

neo4j/test-integration/examples_test.go
robsdedude: If the driver manual tries to pull this in, you'd also have to adjust that (remove the reference from there).

testkit-backend/backend.go
fbiville: actually, you're right, it should work

PR #295, opened 2021-12-17
4.4 tx func flaky tests

fbiville:


fbiville:
I cannot make the suite fail at the moment, so I'll just merge these extra logged info and rely on them whenever the first failure happens again.

PR #302, opened 2022-01-25
Replace `Session.LastBookmark()` with `Session.LastBookmarks()`

robsdedude:
* Returns `[]string` instead of string.
 * Returns all (not just the last) initial bookmarks if no bookmark has been
   received from the server.

Depends on https://github.com/neo4j-drivers/testkit/pull/364


neo4j/session.go
```diff
@@ -35,9 +35,10 @@ type TransactionWork func(tx Transaction) (interface{}, error)
 // Session represents a logical connection (which is not tied to a physical connection)
 // to the server
 type Session interface {
-	// LastBookmark returns the bookmark received following the last successfully completed transaction.
-	// If no bookmark was received or if this transaction was rolled back, the bookmark value will not be changed.
-	LastBookmark() string
+	// LastBookmarks returns the bookmark received following the last successfully completed transaction.
+	// If no bookmark was received or if this transaction was rolled back, the initial set of bookmarks will be
```
fbiville, line 39:
```suggestion
	// If no bookmark were received or if this transaction was rolled back, the initial set of bookmarks will be
```

neo4j/test-integration/bookmark_test.go
```diff
@@ -248,29 +250,29 @@ var _ = Describe("Bookmark", func() {
...
 			Expect(err).To(Not(BeNil()))
 
-			Expect(session.LastBookmark()).To(Equal(bookmark))
+			Expect(session.LastBookmarks()).To(Equal([]string{bookmark}))
 		})
 
 	})
 
-	Context("session constructed with two bookmarks", func() {
+	Context("session constructed with two sets bookmarks", func() {
```
fbiville: what does "two sets bookmarks" mean?

neo4j/util.go
```diff
@@ -0,0 +1,16 @@
+package neo4j
+
+// ConcatenateStringSlices is a helper method to flatten a [][]string slice into a []sting slice.
+// This can be useful for combining bookmarks of multiple sessions.
+func ConcatenateStringSlices(slices... []string) []string  {
```
fbiville: maybe do not export it, i.e. rename it to `concatenateStringSlices`
we don't want users accidentally call that

neo4j/util_test.go
```diff
@@ -0,0 +1,32 @@
+package neo4j
+
+import (
+	"testing"
+	"testing/quick"
+)
+
+func TestConcatenateStringSlices(t *testing.T) {
```
fbiville: nice 😎

neo4j/test-integration/bookmark_test.go
robsdedude: It should be
```suggestion
	Context("session constructed with two sets of bookmarks", func() {
```

neo4j/util.go
robsdedude: Don't we? I thought we'd offer them this method that they might need when working with bookmarks of more than 2 sessions. To avoid them having to type `append(append(s1.LastBookmarks(), s2.LastBookmarks()...), s3.LastBookmarks()...)`. Imagine having to do that for 4 or even more sessions...

neo4j/util_test.go
robsdedude: thanks :sunglasses:

neo4j/util.go
fbiville: ah in that case, it makes sense to rename it to `CombineBookmarks`, as you did initially. Sorry about this 😅

neo4j/util_test.go
fbiville:
```suggestion
func TestCombineBookmarks(t *testing.T) {
```

neo4j/util.go
```diff
@@ -0,0 +1,26 @@
...
+Let s1, s2, s3 be Session interfaces. You can easily causally chain the sessions like so:
+
+	s4 := driver.NewSession(neo4j.SessionConfig{
+		Bookmarks: neo4j.CombineBookmarks(s1.LastBookmarks(), s2.LastBookmarks(), s3.LastBookmarks()),
+	})
+
+The server will then make sure to execute all transactions in s4 after any that were already executed in s1, s2, or s3
+at the time of calling LastBookmarks.
+*/
+func CombineBookmarks(slices... []string) []string  {
```
fbiville: can you rename the file to `bookmarks.go` or `bookmark_util.go` (same for test file)
`util` has the naughty property of acting as a magnet to many unrelated functions

fbiville (approved):

neo4j/util.go
robsdedude: Very good point indeed!

fbiville (approved):

neo4j/bookmarks.go
```diff
@@ -0,0 +1,57 @@
...
+	return res
+}
+
+/*
+BookmarksToRawValues exposes the raw server-side bookmarks.
+
+You should not need to use this method unless you want to serialize bookmarks.
+See Session.LastBookmarks and CombineBookmarks for alternatives.
+ */
+func BookmarksToRawValues(bookmarks Bookmarks) []string {
```
robsdedude, line 45: Ideally, I'd like this to be a receiver function of `Bookmarks`. But since it's a type alias, it can't be one. Maybe deprecate this in 6.0 when `Bookmarks` becomes a real struct and introduce an alternative for this function that is a receiver, then remove this function in 7.0? I hate the massively long scope of this. Any better ideas?

neo4j/bookmarks.go
```diff
@@ -12,15 +23,35 @@ Let s1, s2, s3 be Session interfaces. You can easily causally chain the sessions
...
+	return bookmarks
+}
+
+/*
+BookmarksFromRawValues creates Bookmarks from raw server-side bookmarks.
+
+You should not need to use this method unless you want to de-serialize bookmarks.
+See Session.LastBookmarks and CombineBookmarks for alternatives.
+*/
+func BookmarksFromRawValues(values... string) Bookmarks {
```
fbiville, line 55: nit: `ParseBookmarks` maybe?

neo4j/bookmarks.go
fbiville: since this is only exposed for tests (integration tests and testkit): can we move (and duplicate if needed) this function?

fbiville (approved):

neo4j/bookmarks.go
robsdedude: It's not for tests only. The the docs say, if you want to serialize bookmarks (e.g. to achieve causal consistency between different processes or machines), you need a way to dump/load bookmarks to/fro something easily serializable (e.g. `[]string`).

neo4j/bookmarks.go
robsdedude: Maybe. I was thinking two things: 
 * it should somewhat be obvious that this is the inverse operation to `BookmarksToRawValues` (or whatever we decide to call it)
 * if it starts with "Bookmarks" it will be easier to discover through your IDE's auto-completion.

PR #303, opened 2022-01-26
Remove deprecated elements

fbiville:


neo4j/test-integration/bookmark_test.go
```diff
@@ -346,8 +342,10 @@ var _ = Describe("Bookmark", func() {
 
 			bookmark = createNodeInTx(driver)
 
-			session, err = driver.Session(neo4j.AccessModeWrite, bookmark+"0")
-			Expect(err).To(BeNil())
+			session = driver.NewSession(neo4j.SessionConfig{
```
AndyHeap-NeoTech, line 345: Just want to confirm that you are no longer wanting to check against an error variable being null and that another mechanism is in use.

fbiville: the deprecated `driver.Session` function used to return a session or an error, but the replacement `driver.NewSession` function does not return any error anymore, so there is no error to check

PR #323, opened 2022-02-24
Implement 5.0 handshake

fbiville:


neo4j/dbtype/graph.go
```diff
@@ -22,18 +22,22 @@ package dbtype
 
 // Node represents a node in the neo4j graph database
 type Node struct {
-	Id     int64                  // Id of this node.
-	Labels []string               // Labels attached to this Node.
-	Props  map[string]interface{} // Properties of this Node.
+	Id        int64                  // Id of this Node.
```
bigmontz, line 26: Is there a way of depreciating this Id?

neo4j/internal/bolt/path.go
```diff
@@ -62,16 +63,17 @@ func buildPath(nodes []dbtype.Node, relNodes []*relNode, indexes []int) dbtype.P
...
 		if n1start {
-			rel.StartId = n1.Id
-			rel.EndId = n2.Id
+			rel.StartId, rel.EndId = n1.Id, n2.Id
+			rel.StartElementId, rel.EndElementId = n1.ElementId, n2.ElementId
 		} else {
-			rel.StartId = n2.Id
-			rel.EndId = n1.Id
+			rel.StartId, rel.EndId = n2.Id, n1.Id
+			rel.StartElementId, rel.EndElementId = n2.ElementId, n1.ElementId
```
bigmontz, line 77: I think it could be a good idea add some tests coverage here.

bigmontz (approved):
🔩

neo4j/internal/bolt/path.go
fbiville: there is in `hydrator_test`

neo4j/dbtype/graph.go
fbiville: Yes, but... is it really a deprecation here?
`Id` as `int64` will be gone in 6.0 and replaced by... `Id` as a `string` or am I missing something?

bigmontz: For changing the type safely in `6.0`, is it not important making devs do not use the property as an integer first?

fbiville: The problem I see is two-fold:

 - `Id` will be deprecated without a proper alternative **beyond** 5.x - in 6.0, there will be a breaking change no matter what
 - shouldn't `ElementId` be deprecated already since it's gonna be renamed in 6.0?
 
I think deprecations will bring more confusion. 
I agree, however, that I should explicitly document the future plans for `Id` and `ElementId` in the docs.

fbiville: My bad, `ElementId` is here to stay.. Let me add proper docs.

fbiville: fixed in c3df8b9a524be38d11a81a119070f58d63a068c3

PR #325, opened 2022-02-25
Do not retry transactions on authentication errors

fbiville:


bigmontz (approved):
🔑 

(It may could worth add an unit test)

fbiville:
Done 📈

PR #334, opened 2022-03-24
Fix summary test flakiness, init timers with negative value

fbiville:


bigmontz (approved):
🧠 

You could may worth add some docs around the values, just to make explicitly the meaning of the `-1` since it was not present in the 4.4 version.

fbiville:
> You could may worth add some docs around the values, just to make explicitly the meaning of the `-1` since it was not present in the 4.4 version.

✅

PR #335, opened 2022-03-24
Unify transaction behavior with other drivers

fbiville:
There is now no need to explicit roll back a failed transaction
(run).

Commit and Rollback can only be called once.

Rollback does not return an error from a failed run anymore.


fbiville:
Companion PR: https://github.com/neo4j-drivers/testkit/pull/438

AndyHeap-NeoTech (approved):
LGTM

PR #341, opened 2022-03-29
Minimize sent RESETs by removing ForceReset internal API

fbiville:


robsdedude (approved):
:stop_sign:

neo4j/internal/bolt/bolt3.go
```diff
@@ -668,6 +668,10 @@ func (b *bolt3) Birthdate() time.Time {
 }
 
 func (b *bolt3) Reset(ctx context.Context) {
+	if b.state == bolt3_ready {
```
fbiville: need to revisit these, it's not a good idea to not reset the internal state of ready connections

fbiville: we can go back to ready when a tx is committed e.g., in which case we want to reset the internal state before reusing the connection

PR #351, opened 2022-04-13
Expose official API to determine whether errors are retryable

fbiville:


neo4j/internal/retry/state.go
```diff
@@ -115,18 +116,16 @@ func (s *State) OnFailure(conn idb.Connection, err error, isCommitting bool) {
 		return
 	}
 
+	s.LastErrWasRetryable = IsRetryable(err)
```
bigmontz, line 119: How go driver know it should retry on `ConnectivityError`?

bigmontz (approved):
🦜

neo4j/internal/retry/state.go
fbiville: `ConnectivityError` is higher-level than what the driver internally deals with.

If you look at the [retry code](https://github.com/neo4j/neo4j-go-driver/blob/4.4/neo4j/internal/retry/state.go#L85-L107), the driver deals with the internal `Connection` object and knows:

- if the pool borrow failed (connection is `nil`), this will lead to a `ConnectivityError`
- if the connection is in a dead state (which would result in higher-level API in a `ConnectivityError`)

PR #352, opened 2022-04-15
Implement GetServerInfo

fbiville:


fbiville:
I'll split the work in two PRs:

1. implementing `GetServerInfo` here
2. "rebase" `VerifyConnectivity` on top of it in a second PR

neo4j/internal/bolt/bolt3_test.go
```diff
@@ -512,4 +513,39 @@ func TestBolt3(ot *testing.T) {
 		AssertNil(t, sum)
 		AssertError(t, err)
 	})
+
+	outer.Run("Updates idle date on every new message", func(t *testing.T) {
```
robsdedude:
```suggestion
	outer.Run("Updates idle date on SUCCESS after RESET", func(t *testing.T) {
```

It's not really testing every message.

neo4j/internal/bolt/bolt3_test.go
```diff
@@ -512,4 +513,39 @@ func TestBolt3(ot *testing.T) {
...
+		defer bolt.Close(ctx)
+
+		firstIdleDate := bolt.IdleDate()
+		AssertAfter(t, firstIdleDate, testStart)
+
+		bolt.ForceReset(ctx)
+		AssertAfter(t, bolt.IdleDate(), firstIdleDate)
+	})
+
+	outer.Run("Does not update idle date on error", func(t *testing.T) {
```
robsdedude:
```suggestion
	outer.Run("Does not update idle date on disconnect", func(t *testing.T) {
```
I'd expect it to update the idle date on receiving `FAILURE`.

neo4j/internal/bolt/bolt4_test.go
```diff
@@ -977,4 +977,39 @@ func TestBolt4(ot *testing.T) {
 		assertBoltState(t, bolt4_failed, bolt)
 		AssertError(t, err)
 	})
+
+	outer.Run("Updates idle date on every new message", func(t *testing.T) {
```
robsdedude: s. above

neo4j/internal/bolt/bolt4_test.go
```diff
@@ -977,4 +977,39 @@ func TestBolt4(ot *testing.T) {
...
+		defer bolt.Close(ctx)
+
+		firstIdleDate := bolt.IdleDate()
+		AssertAfter(t, firstIdleDate, testStart)
+
+		bolt.ForceReset(ctx)
+		AssertAfter(t, bolt.IdleDate(), firstIdleDate)
+	})
+
+	outer.Run("Does not update idle date on error", func(t *testing.T) {
```
robsdedude: s. above

neo4j/internal/bolt/bolt5_test.go
```diff
@@ -924,4 +924,39 @@ func TestBolt5(ot *testing.T) {
 		assertBoltState(t, bolt5Failed, bolt)
 		AssertError(t, err)
 	})
+
+	outer.Run("Updates idle date on every new message", func(t *testing.T) {
```
robsdedude: s. above

neo4j/internal/bolt/bolt5_test.go
```diff
@@ -924,4 +924,39 @@ func TestBolt5(ot *testing.T) {
...
+		defer bolt.Close(ctx)
+
+		firstIdleDate := bolt.IdleDate()
+		AssertAfter(t, firstIdleDate, testStart)
+
+		bolt.ForceReset(ctx)
+		AssertAfter(t, bolt.IdleDate(), firstIdleDate)
+	})
+
+	outer.Run("Does not update idle date on error", func(t *testing.T) {
```
robsdedude: s. above

neo4j/internal/pool/pool.go
```diff
@@ -29,13 +29,18 @@ import (
...
 	"sort"
 	"sync"
 	"time"
 
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/log"
 )
 
+// DefaultLivenessCheckThreshold disables the liveness check of connections
+// Liveness checks are performed before a connection is deemed idle enough to be reset
+const DefaultLivenessCheckThreshold = math.MaxInt64
```
robsdedude, line 42: Why not choosing `-1` (or any negative value) to signal no need for a liveness check? That way, MaxInt64 is still a valid (albeit questionable) value for configuring the timeout.

neo4j/internal/pool/server.go
```diff
@@ -41,17 +41,21 @@ var sharedRoundRobin uint32
 
 const rememberFailedConnectDuration = 3 * time.Minute
 
-// Returns a idle connection if any
-func (s *server) getIdle() db.Connection {
+// Returns an idle connection if any
+func (s *server) getIdle(ctx context.Context, idlenessThreshold time.Duration) db.Connection {
```
robsdedude: Do the looping over idle connections, waiting for the result of the forced reset. https://neo4j.slack.com/archives/D01MXVBHT26/p1651766602894969

neo4j/internal/pool/server_test.go
```diff
@@ -106,23 +106,23 @@ func TestServer(ot *testing.T) {
...
 
-		// Let the conn in the middle be too old
+		// Let the connection in the middle be too old
 		conns[1].Birth = now.Add(-20 * time.Second)
 		s.removeIdleOlderThan(context.Background(), now, 10*time.Second)
 		assertSize(t, s, 2)
 
 		// Should be able to borrow twice
-		b1 := s.getIdle()
+		b1 := s.getIdle(nil, 0)
```
robsdedude: should this be `context.Background()` instead of `nil` for style-points or does it not really matter in tests?

(many more like this, which I will not comment on)

neo4j/internal/router/router.go
```diff
@@ -52,7 +52,10 @@ type Router struct {
 }
 
 type Pool interface {
-	Borrow(ctx context.Context, servers []string, wait bool, boltLogger log.BoltLogger) (db.Connection, error)
+	// Borrow acquires a connection from the provided list of servers
+	// If all connections are busy and the pool is full, calls to Borrow may wait for a connection to become idle
+	// If a connection has been idle for enough time, based on the idlenessThreshold, it will be reset.
```
robsdedude:
```suggestion
	// If a connection has been idle for longer than idlenessThreshold, it will be reset
	// to check if it's still alive.
```

neo4j/internal/testutil/poolfake.go
```diff
@@ -32,17 +33,17 @@ type PoolFake struct {
...
 
-func (p *PoolFake) Return(ctx context.Context, c db.Connection) {
+func (p *PoolFake) Return(context.Context, db.Connection) {
 	if p.ReturnHook != nil {
 		p.ReturnHook()
 	}
 }
 
-func (p *PoolFake) CleanUp(ctx context.Context) {
+func (p *PoolFake) CleanUp(context.Context) {
```
robsdedude, line 53: In this PR, I've unnamed found unused arguments (like here) and `_`. which one is the preferred style?

neo4j/session_with_context_test.go
```diff
@@ -569,7 +569,74 @@ func TestSession(st *testing.T) {
 		})
 	})
 
-	st.Run("Close", func(ct *testing.T) {
+	outer.Run("GetServerInfo", func(inner *testing.T) {
+
+		inner.Run("Retrieves info from first successful connection", func(t *testing.T) {
```
robsdedude: It's not really testing it it's the first successful connection as the fake pool will return the same fake connection as many times as asked to.

testkit-backend/backend.go
```diff
@@ -437,10 +437,26 @@ func (b *backend) handleRequest(req map[string]interface{}) {
...
+			"agent":           serverInfo.Agent(),
+			"protocolVersion": fmt.Sprintf("%d.%d", protocolVersion.Major, protocolVersion.Minor),
+		})
+
 	case "NewSession":
 		driver := b.drivers[data["driverId"].(string)]
-		sessionConfig := neo4j.SessionConfig{}
-		sessionConfig.BoltLogger = neo4j.ConsoleBoltLogger()
+		sessionConfig := neo4j.SessionConfig{
+			BoltLogger: neo4j.ConsoleBoltLogger(),
```
robsdedude, line 458: :tada:

robsdedude (changes requested):

neo4j/internal/db/connection.go
```diff
@@ -102,7 +102,7 @@ type Connection interface {
 	// Active streams will be discarded and the bookmark will be lost.
 	Reset(ctx context.Context)
 	// ForceReset is like Reset except it also sends a RESET messages for ready
-	// connections
+	// connections too
```
robsdedude:
```Go
	// ForceReset is like Reset except it also sends a RESET message for ready
	// connections, too.
```
Maybe? (punctuation and message in singular form)

neo4j/internal/bolt/bolt3_test.go
fbiville: true, but the test implementation shows this, while the intention and the implementation shows it's done on every message

fbiville: ideally, the test would be more general and indeed test any message but that would be cumbersome to set up.
I'd like to keep the intention in the test name. Thoughts?

neo4j/internal/testutil/poolfake.go
fbiville: `_` is used when there are named parameters
if none of the parameters are named, names can be scratched completely (have you seen an instance where all of the names were `_`?)

neo4j/session_with_context_test.go
fbiville: I can make the test a bit stricter by introducing `BorrowHook`

neo4j/internal/db/connection.go
fbiville: rephrased

neo4j/session_with_context_test.go
```diff
@@ -569,7 +570,82 @@ func TestSession(st *testing.T) {
 		})
 	})
 
-	st.Run("Close", func(ct *testing.T) {
+	outer.Run("GetServerInfo", func(inner *testing.T) {
+
+		inner.Run("Retrieves info from first borrowed connection", func(t *testing.T) {
```
bigmontz, line 575: We could test also the case where the connection already exists.

bigmontz (approved):
🌵

neo4j/internal/bolt/bolt3_test.go
bigmontz: Is there no easy way of setup a scenario for all messages? Like using the two functions per scenario, one to setup the mocks for the sever and other to call the method, this way you could use a parameterized test with pair of function for each message you wan't to check.

thelonelyvulpes (approved):

neo4j/internal/pool/pool.go
fbiville: that would make the comparison a bit less obvious
the comparison is currently:
```go
if time.Now().Sub(connection.IdleDate()) > idlenessThreshold {
	connection.ForceReset(ctx)
	if !connection.IsAlive() {
		return nil, found
	}
}
```
with the current default, a simple comparison is enough to be sure that we won't reset idle connections

neo4j/internal/testutil/poolfake.go
fbiville: I just checked, this PR only uses `_` to ignored part of the returned results or part of the parameters, which is the way go...
when all parameters are ignored like above, the name part is completely omitted

neo4j/session_with_context_test.go
fbiville: since we mock the entire pool here, we cannot really make that distinction.
it's at the pool level that we can test such things, not at the session level

neo4j/internal/bolt/bolt3_test.go
fbiville: updated with parameterizated test cases, covering almost every message

neo4j/internal/bolt/bolt3_test.go
fbiville: since this is a real TCP server under the hood, it's kinda hard to produce anything else than a disconnect error.
However, I parameterized *when* the error happens, by including those cases into the... matrix.

bigmontz (approved):
🥥

PR #353, opened 2022-04-20
Remove lazy begin state tracking from Bolt 5 connections

fbiville:
This was already removed from Bolt 3 and 4 connections.


robsdedude (approved):
:running_shirt_with_sash:

PR #359, opened 2022-05-09
Do not log in unit(-ish) tests

fbiville:
Assertions should always be a good enough indicator of what is
going wrong


AndyHeap-NeoTech (approved):
Makes sense to me.

PR #363, opened 2022-05-12
Introduce context-aware lock

fbiville:


neo4j/internal/racingio/lock.go
```diff
@@ -0,0 +1,44 @@
...
+	case c.ch <- struct{}{}:
+		return true
+	case <-ctx.Done():
+		return false
+	}
+}
+
+func (c *contextLock) Unlock() {
+	<-c.ch
+}
```
robsdedude: Does this mean that calling `Unlock` on a Mutex that is not locked will deadlock your program? What about 
```suggestion
func (c *contextLock) Unlock() {
	select {
	case <-c.ch:
		return
	default:
		panic("Cannot unlock unlocked mutex.")
    }
}
```

neo4j/internal/racingio/lock_test.go
```diff
@@ -0,0 +1,51 @@
...
+	outer.Run("fails to lock after deadline reached", func(t *testing.T) {
+		delay := 20 * time.Millisecond
+		timeout, cancelFunc := context.WithTimeout(backgroundCtx, delay)
+		defer cancelFunc()
+		mutex := racingio.NewMutex()
+		time.Sleep(2 * delay)
+
+		result := mutex.TryLock(timeout)
+
+		testutil.AssertFalse(t, result)
```
robsdedude: What about adding a case where the mutex is already locked and a goroutine tries to acquire it while the main routine releases if but after the timeout?

fbiville: Good idea, added now :)

neo4j/internal/racingio/lock.go
fbiville: I've not given too much thought about `Unlock` just yet.
I'll probably return `false` in case it's not locked, to provide symmetry with `Lock` (`panic` is rarely used).

robsdedude (approved):
Maybe consider a panic when the user attempts to unlock an unlocked mutex (like the [Go built-in mutex does](https://pkg.go.dev/sync#RWMutex.Unlock)) as that's a state you're not supposed to be in (like trying to access a variable before defining it).

But since it's an internal lib, you can also do it this way. Whatever you prefer. The rest of the PR LGTM.

neo4j/internal/pool/pool.go
```diff
@@ -256,7 +287,10 @@ func (p *Pool) Borrow(ctx context.Context, serverNames []string, wait bool, bolt
 	case <-q.wakeup:
 		return q.conn, nil
 	case <-ctx.Done():
-		p.queueMut.Lock()
+		// TODO: provided ctx has reached deadline already - set some hardcoded timeout instead?
+		if !p.queueMut.TryLock(context.Background()) {
```
fbiville, line 292: this emulates the previous implementation which could possibly block forever... not sure that's the best thing to do, though.

neo4j/internal/racing/lock.go
```diff
@@ -30,14 +30,16 @@ func NewMutex() Mutex {
...
+	deadline, hasDeadline := ctx.Deadline()
+	err := ctx.Err()
+	switch {
+	case !hasDeadline && err == nil:
 		c.ch <- struct{}{}
 		return true
-	} else if deadline.Before(time.Now()) {
-		// if the context is already done and the lock is available
-		// select will pick any of the case at random
+	case deadline.Before(time.Now()) || err != nil:
```
robsdedude, line 39: Is there a potential race-condition? What if, at this point, the deadline has not yet expired, but it does between this check the the following `select` block? I'm wondering, is the `.Before` check even necessary? What if the user wants to try to acquire the lock without waiting (timeout 0 seconds). What context would they chose?

This might be controversial but I'd suggest that on expired contexts, we try to acquire the lock and if we can't do so immediately, we return `false`. Or in other words a non-blocking lock acquisition.

neo4j/internal/pool/pool.go
```diff
@@ -232,16 +258,21 @@ func (p *Pool) Borrow(ctx context.Context, serverNames []string, wait bool, bolt
...
+	if !p.queueMut.TryLock(ctx) {
+		return nil, racing.LockTimeoutError("could not acquire lock in time when trying to get an idle connection")
+	}
 	// Ok, now that we own the queue we can add the item there but between getting the lock
 	// and above check for an existing connection another thread might have returned a connection
 	// so check again to avoid potentially starving this thread.
-	conn = p.tryAnyIdle(ctx, serverNames, idlenessThreshold)
+	conn, err = p.tryAnyIdle(ctx, serverNames, idlenessThreshold)
+	if err != nil {
+		return nil, err
```
robsdedude, line 270: There is no release between the lock acquisition and this return => deadlock potential?

neo4j/internal/racing/lock.go
fbiville: I don't think it's a race condition in that case.
If we're unlucky enough to "escape" the guard here but the context expires right before `select` is evaluated, then the context has expired and the lock should not be acquired.

neo4j/internal/pool/pool.go
fbiville: oopsie, good catch, thanks :)

neo4j/internal/racing/lock.go
fbiville: To elaborate: the sole purpose of this early check is to avoid the situation where the lock is available and the context has expired at the same time. In that specific case, without this early check, `select` would pick at random, which is not what we want. If both the lock is available and the context has expired, we don't want to acquire the lock.

If this check passes, it does not necessarily mean the lock will be acquired, as you pointed out. But I don't think it's an issue.

robsdedude: Fair enough. But the random choice can still happen iff the first check passes and the context times out right afterwards and the mutex is free.

fbiville: true, but the window becomes much smaller so I think it's acceptable

PR #364, opened 2022-05-18
Warn Config#TlsConfig "voids warranties"

fbiville:


AndyHeap-NeoTech (approved):
As concise and clear a warning as we can do.

PR #367, opened 2022-05-19
[4.3] Handle transaction function panics

fbiville:


AndyHeap-NeoTech (approved):
Looks ok to me. One question because of my ignorance of Go. Is it normal pattern to have a single function contain multiple tests in the world of Go?

fbiville:
> Looks ok to me. One question because of my ignorance of Go. Is it normal pattern to have a single function contain multiple tests in the world of Go?

Yep, that's the way to have nested tests.

PR #369, opened 2022-05-24
Test speedup

fbiville:


neo4j/internal/bolt/dechunker_test.go
```diff
@@ -42,10 +42,10 @@ func TestDechunker(t *testing.T) {
 	}{
 		{size: 3, max: 0xffff},
 		{size: 1021, max: 0x7},
-		{size: 0xffff78, max: 0x30},
+		{size: 0xffff, max: 0x30},
 		{size: 3, max: 0xffff},
 		{size: 1021, max: 0xff90},
-		{size: 0xffff78, max: 0xff90},
```
robsdedude, line 48: Why was it that odd (not in a mathematical sense) number? And why is it fine to lower the expected range in which the code works?

fbiville: I double-checked the implementation and could not see why that particular message size mattered.
I think the important bit to test is that some test payloads need to be read into several chunks.
Aiming for a big payload like this one is just too slow (the whole test took 5+ seconds, which becomes quite noticeable when running unit tests).

robsdedude (approved):
:runner:

PR #370, opened 2022-05-25
Reclassify specific 4.x errors into their 5.x classifications

fbiville:


neo4j/db/errors.go
```diff
@@ -73,6 +74,18 @@ func (e *Neo4jError) parse() {
...
+// this function can be removed once support for pre-5.x servers is dropped
+func (e *Neo4jError) reclassify() {
+	switch e.Code {
+	case "Neo.TransientError.Transaction.LockClientStopped":
+		e.Code = "Neo.ClientError.Transaction.LockClientStopped"
+	case "Neo.TransientError.Transaction.Terminated":
+		e.Code = "Neo.ClientError.Transaction.Terminated"
+	}
+}
+
```
bigmontz, line 88: Why should we reclassify the error in the driver code?

fbiville: I also think this should be done on the server side but that's water under the bridge.
The decision is that drivers are responsible for mapping 4.x errors into their 5.x classifications since retries are a driver-only concern.

bigmontz (approved):
🥓

neo4j/db/errors.go
robsdedude: It is just not possible to change the already released 4.x servers. Hence this is necessary. Maybe you want to add a to do comment along the lines of "remove when 4.x support gets removed". Ofc, that only helps when you periodically check the to do comments.

robsdedude (approved):
:arrow_right: :writing_hand:

fbiville:
Now the fun part begins: builds will have to fail before everything resolves again (once both the driver and testkit PR are merged).

PR #383, opened 2022-07-25
Support UTC DateTime Structures

fbiville:
Contrary to 5.0, invalid timezones received from the server will
result in the transaction being retried and cause the connection
to eventually put back to pool, after all attempts are
exhausted.

While this is not ideal, this has been done so to minimize
breaking changes in 4.x.


dotStart:
[Branch: NEO4J_DEV
[Veracode Flaw 75b94708] Module: jetty-http-9.4.44.v20210927.jar Severity score: 4.0](https://trello.com/c/QGVdCVZu/4432-branch-neo4jdev-veracode-flaw-75b94708-module-jetty-http-9444v20210927jar-severity-score-40)

PR #384, opened 2022-07-26
Support UTC DateTime Structures

fbiville:
Contrary to 5.0, invalid timezones received from the server will
result in the transaction being retried and cause the connection
to eventually put back to pool, after all attempts are
exhausted.

While this is not ideal, this has been done so to minimize
breaking changes in 4.x.


neo4j/test-integration/dbconn_test.go
```diff
@@ -236,7 +234,7 @@ func TestConnectionConformance(ot *testing.T) {
 		{
 			name: "Next passed the summary",
 			fun: func(t *testing.T, c db.Connection) {
-				s, err := boltConn.Run(db.Command{Cypher: "RETURN datetime()"}, db.TxConfig{Mode: db.ReadMode})
+				s, err := boltConn.Run(db.Command{Cypher: "RETURN 42"}, db.TxConfig{Mode: db.ReadMode})
```
bigmontz, line 237: Why did this test change?

bigmontz (approved):
🚀

neo4j/test-integration/dbconn_test.go
fbiville: because the driver would send a UTC DateTime structure and the server would not understand it (yet).
Since that test does not really need any DateTime to make its point, I decided to switch to a simpler value.

PR #388, opened 2022-08-04
Bookmark manager

fbiville:
⚠️ depends on https://github.com/neo4j-drivers/testkit/pull/506 ⚠️

### DIFF
--- a/neo4j/bookmarks.go
+++ b/neo4j/bookmarks.go
@@ -1,4 +1,28 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [https://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package neo4j
+
+import (
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/collection"
+	"sync"
+)
 
 // Bookmarks is a holder for server-side bookmarks which are used for causally-chained sessions.
 // See also CombineBookmarks.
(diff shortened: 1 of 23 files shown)


neo4j/bookmarks.go
```diff
@@ -10,13 +35,129 @@ type BookmarkManager interface {
...
+		//TODO: better error handling?
+		panic("could not acquire write lock for updated database bookmarks")
+	}
+	defer databaseLock.UnlockWrite()
+
+	currentBookmarks, found := b.bookmarks[database]
+	if !found {
+		b.bookmarks[database] = collection.NewSet(newBookmarks)
+		return
+	}
```
bigmontz: You should notify the current bookmarks also here.

fbiville: obrigado

fbiville: Fixed in e628a710dceaa92357a96caac64e4768efbc30da

neo4j/internal/bolt/bolt3.go
```diff
@@ -617,8 +617,8 @@ func (b *bolt3) receiveNext(ctx context.Context) (*db.Record, *db.Summary, error
 	}
 }
 
-func (b *bolt3) Bookmark() string {
-	return b.bookmark
+func (b *bolt3) Bookmark() (string, string) {
+	return b.bookmark, ""
 }
```
bigmontz: We may need some docs here. I took some time to figure out the return from this method.

neo4j/internal/db/connection.go
```diff
@@ -79,11 +79,12 @@ type Connection interface {
 	// streams are created and the server doesn't support multiple streams. Use Buffer to force
 	// buffering before calling Reset to get all records and the bookmark.
 	Buffer(ctx context.Context, streamHandle StreamHandle) error
-	// Bookmark returns the bookmark from last committed transaction or last finished auto-commit transaction.
+	// Bookmark returns the bookmark and optionally its database from last committed transaction or last finished auto-commit transaction.
+	// The database is only populated if queries are executed with the USE clause.
```
bigmontz: The database is also populated without the usage of the `USE` cause, but in this case, it will be equal the one we have in the session.

bigmontz (approved):
🔖 🧑‍💼

neo4j/internal/bolt/bolt3.go
fbiville: It's documented at the interface level: https://github.com/neo4j/neo4j-go-driver/blob/9cd76fe4dc146d92df981c2bae0eb5e20628a5e0/neo4j/internal/db/connection.go#L82-L87
Do you think I should add a specific comment for that empty string here? Or sth else?

neo4j/internal/db/connection.go
fbiville: clarified in 112731a6f954224f93cfa92771657246abcbea04 (hopefully)

neo4j/internal/bolt/bolt3.go
bigmontz: It may worth it since we were not that close to the context where the interface is defined.

fbiville: 77d53e34897bf4e9c3ab754a60e59f385935d6e8

PR #390, opened 2022-08-09
Extract current bookmark management into separate struct

fbiville:


bigmontz (approved):
🔖

robsdedude (approved):
I guess the unit tests could be a bit more comprehensive. But any tests are better than the 0 tests we had for that concern before. :heavy_check_mark:

PR #391, opened 2022-08-09
Retrieve bookmarks after autocommit result consumption

fbiville:
... instead of retrieving them when Session#LastBookmark(s) is
called since there is no guarantee LastBookmark(s) has been called
after the results were consumed.

### DIFF
--- a/neo4j/result.go
+++ b/neo4j/result.go
@@ -21,7 +21,6 @@ package neo4j
 
 import (
 	"context"
-	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/db"
 )
 
 // Deprecated: use ResultWithContext instead.
@@ -54,11 +53,6 @@ type Result interface {
 // deprecated: use resultWithContext instead
 type result struct {
 	delegate ResultWithContext
-}
-
-// deprecated: use newResultWithContext instead
-func newResult(conn db.Connection, str db.StreamHandle, cypher string, params map[string]any) Result {
-	return newResultWithContext(conn, str, cypher, params).legacy()
 }
 
 func (r *result) Keys() ([]string, error) {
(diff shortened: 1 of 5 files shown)


neo4j/result.go
```diff
@@ -56,11 +55,6 @@ type result struct {
 	delegate ResultWithContext
 }
 
-// deprecated: use newResultWithContext instead
-func newResult(conn db.Connection, str db.StreamHandle, cypher string, params map[string]any) Result {
-	return newResultWithContext(conn, str, cypher, params).legacy()
-}
-
```
bigmontz, line 63: Is it accessible from the public api?

bigmontz (approved):
🚧

neo4j/result.go
fbiville: nope (since it starts with a lower case letter, it's not visible outside of the package) and it was actually not used anymore

neo4j/result_with_context.go
```diff
@@ -60,25 +60,26 @@ type ResultWithContext interface {
...
+	streamHandle                   idb.StreamHandle
+	cypher                         string
+	params                         map[string]any
+	record                         *Record
+	summary                        *db.Summary
+	err                            error
+	peekedRecord                   *Record
+	peekedSummary                  *db.Summary
+	peeked                         bool
+	afterAutocommitConsumptionHook func()
```
robsdedude: I think the name might be misleading. Isn't the cb called whenever the result is exhausted, not just for auto-commits?

robsdedude (approved):

neo4j/result_with_context.go
fbiville: yes, but it's only non-nil for autocommit (since that's the only situations where bookmarks are returned after the result consumption)

fbiville: Renamed in a more generic way

PR #396, opened 2022-08-25
Move BookmarkManager at the session level

fbiville:
Temporarily remove TestKit integration


fbiville:
Do not merge before https://github.com/neo4j-drivers/testkit/pull/509

PR #398, opened 2022-09-07
Database unavailable

fbiville:


neo4j/internal/retry/state.go
```diff
@@ -134,6 +136,19 @@ func (s *State) OnFailure(ctx context.Context, conn idb.Connection, err error, i
...
+			invalidationFn := s.Router.InvalidateWriter
+			if mode == idb.ReadMode {
+				invalidationFn = s.Router.InvalidateReader
+			}
+			if err := invalidationFn(ctx, s.DatabaseName, conn.ServerName()); err != nil {
+				s.stop = true
+				s.LastErr = err
+			}
+			s.cause = "Database unavailable error"
+			return
```
bigmontz: if a `conn` to the router arrives here, should it be remove from write/read or just invalidate the server name?

fbiville: can you get a `Neo.TransientError.General.DatabaseUnavailable` with a router connection? 🤔

fbiville: moreover, this snippet here is called when transaction function failures occur, so a router connection would not be used here, but rather a reader or a writer one, or?

PR #404, opened 2022-10-11
Update manual with 5.0 examples

fbiville:


neo4j/result_helpers.go
```diff
@@ -78,19 +79,32 @@ func Single(result Result, err error) (*Record, error) {
...
+// Deprecated: use CollectWithContext instead (the entry point of context-aware
+// APIs is NewDriverWithContext)
 func Collect(result Result, err error) ([]*Record, error) {
 	if err != nil {
 		return nil, err
 	}
 	return result.Collect()
 }
 
+// CollectWithContext behaves the same as Collect
```
robsdedude: Maybe move the main documentation here as you want users who use the new method to have an easy time.
Instead refer to the docs of `CollectWithContext` in the docs of `Collect`.

robsdedude (approved):
Well contextualized :ok_hand:

neo4j/result_helpers.go
fbiville: Fixed in 56dc354854459b43e676cbc21baaf457ac54d477

PR #407, opened 2022-10-20
Implement query execution API at the driver level

fbiville:


neo4j/driver_with_context.go
```diff
@@ -48,6 +50,48 @@ const (
...
+	//
+	// This function runs the query in a single explicit transaction within a session, entirely managed by the driver.
+	//
+	// Specific settings can be configured via configuration callbacks. Built-in callbacks are provided such as:
+	//	neo4j.WithDatabase
+	//	neo4j.WithWritersRouting
+	//  ...
+	// see neo4j.ExecuteQueryConfiguration for all possibilities.
+	//
+	// These built-in can be used and combined as follows:
```
robsdedude:
```suggestion
	// These built-in callbacks can be used and combined as follows:
```

neo4j/driver_with_context.go
```diff
@@ -48,6 +50,48 @@ const (
 // DriverWithContext represents a pool(s) of connections to a neo4j server or cluster. It's
 // safe for concurrent use.
 type DriverWithContext interface {
+	// ExecuteQuery runs the specified query with its parameters and returns results.
+	// 	results, err := driver.ExecuteQuery(ctx, query, params)
+	//
+	// This function runs the query in a single explicit transaction within a session, entirely managed by the driver.
```
robsdedude: * doesn't mention potential retires.
 * explicit transaction in the Python API docs refers to `session.begin_transaction` i.e., unmanaged transactions. Not sure if this could cause confusion.

neo4j/driver_with_context.go
```diff
@@ -48,6 +50,48 @@ const (
...
+	// This is useful when ExecuteQuery is called without custom bookmark managers and the lower-level
+	// neo4j.SessionWithContext APIs are called as well.
+	// In that case, the recommended approach is as follows:
+	// 	results, err := driver.ExecuteQuery(ctx, query, params)
+	// 	// [...] do something with results and error
+	//	bookmarkManager := driver.GetDefaultManagedBookmarkManager()
+	// 	// maintain consistency with sessions as well
+	//	session := driver.NewSession(ctx, neo4j.SessionConfig {BookmarkManager: bookmarkManager})
+	//	// [...] run something within the session
+	GetDefaultManagedBookmarkManager() BookmarkManager
```
robsdedude: Interestingly enough, the ADR doesn't specify the name of this getter. Java, JS, and Python went with `queryBookmarkManager`, while .NET seems to not expose it for now. Before merging any of the languages, we should sit together one more time and discuss the naming.

neo4j/driver_with_context.go
```diff
@@ -48,6 +50,51 @@ const (
 // DriverWithContext represents a pool(s) of connections to a neo4j server or cluster. It's
 // safe for concurrent use.
 type DriverWithContext interface {
+	// ExecuteQuery runs the specified query with its parameters and returns results.
+	// 	results, err := driver.ExecuteQuery(ctx, query, params)
+	//
+	// This function runs the query in a single explicit, retryable transaction within a session, entirely managed by
+	// the driver.
+	// Because it is an explicit transaction, Cypher queries using "CALL {} IN TRANSACTIONS" or the older
+	// "USING PERIODIC COMMIT" construct will not work (call SessionWithContext#Run for these).
```
robsdedude: Good point that other API docs should probably pick up as well. :clap:

neo4j/driver_with_context.go
```diff
@@ -48,6 +50,51 @@ const (
...
+	//
+	// Specific settings can be configured via configuration callbacks. Built-in callbacks are provided such as:
+	//	neo4j.WithDatabase
+	//	neo4j.WithWritersRouting
+	//  ...
+	// see neo4j.ExecuteQueryConfiguration for all possibilities.
+	//
+	// These built-in can be used and combined as follows:
```
robsdedude:
```suggestion
	// These built-in callbacks can be used and combined as follows:
```

neo4j/driver_with_context.go
```diff
@@ -315,3 +369,144 @@ func (d *driverWithContext) Close(ctx context.Context) error {
...
+	}
+}
+
+func WithImpersonatedUser(user string) ExecuteQueryConfigurationOption {
+	return func(configuration *ExecuteQueryConfiguration) {
+		configuration.ImpersonatedUser = user
+	}
+}
+
+func WithDatabase(db string) ExecuteQueryConfigurationOption {
```
robsdedude: Does this one or any of the others have the potential to collide with other such helper functions we might decide to introduce in the future, e.g., to change how the session is configured?

neo4j/driver_with_context.go
```diff
@@ -315,3 +369,144 @@ func (d *driverWithContext) Close(ctx context.Context) error {
...
+		return session.ExecuteRead, nil
+	case Writers:
+		return session.ExecuteWrite, nil
+	}
+	return nil, fmt.Errorf("unsupported routing control, got: %d", c.Routing)
+}
+
+type EagerResult struct {
+	Keys    []string
+	Records []*Record
```
robsdedude, line 558: Out of curiosity: why are records pointers being passed around? Is that for a lower memory and runtime profile?

neo4j/driver_with_context_examples_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"context"
+	"fmt"
+)
+
+var myDriver DriverWithContext
+var ctx context.Context
+
+func ExampleDriverWithContext_ExecuteQuery() {
+	results, err := myDriver.ExecuteQuery(ctx, "RETURN $value AS val", map[string]any{"value": 42})
```
robsdedude:
```suggestion
	result, err := myDriver.ExecuteQuery(ctx, "RETURN $value AS val", map[string]any{"value": 42})
```
maybe. As it's a single `EagerResult`.

neo4j/driver_with_context_examples_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+var myDriver DriverWithContext
+var ctx context.Context
+
+func ExampleDriverWithContext_ExecuteQuery() {
```
robsdedude, line 11: I don't fully understand how this test works, and what it's there for. I don't see any assertions of something.

neo4j/driver_with_context_examples_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// the following transaction function is guaranteed to see the result of the previous query
+	// since the session uses the same bookmark manager as the previous ExecuteQuery call and targets the same
+	// (default) database
+	count, err := session.ExecuteRead(ctx, func(tx ManagedTransaction) (any, error) {
+		results, err := tx.Run(ctx, "MATCH (n:Example) RETURN count(n) AS count", nil)
+		if err != nil {
+			return nil, err
+		}
+		record, err := results.Single(ctx)
```
robsdedude: General feedback (regardless of this PR): I wan't sure if `Result.Single` would consume the result. The docs weren't clear about it either so I had to check the source code. Maybe extend the docs at some point.

neo4j/driver_with_context_test.go
```diff
@@ -0,0 +1,507 @@
...
+				Keys:    keys,
+				Records: records,
+				Summary: summary,
+			},
+		},
+		{
+			description: "returns error when write result keys cannot be retrieved",
+			createSession: &fakeSession{
+				executeWriteTransactionResult: &fakeResult{
+					keysErr: fmt.Errorf("dude, where are my keys"),
```
robsdedude, line 136: I feel seen :see_no_evil:

neo4j/driver_with_context_test.go
```diff
@@ -0,0 +1,507 @@
...
+		}
+		address := uintptr(unsafe.Pointer(driver.defaultManagedSessionBookmarkManager.(*bookmarkManager)))
+		if count, found := addressCounts[address]; !found || count != int32(goroutineCount) {
+			t.Errorf("expected pointer address %v to be seen %d time(s), got these instead %v", address, count, addressCounts)
+		}
+	})
+}
+
+func callExecuteQueryOrBookmarkManagerGetter(driver *driverWithContext, i int) {
+	if i%2 == 0 {
```
robsdedude, line 324:
```suggestion
	if i % 2 == 0 {
```
maybe? don't know much about Go style guide but looked off to me.

neo4j/driver_with_context_test.go
```diff
@@ -0,0 +1,507 @@
...
+			},
+		}
+
+		var wait sync.WaitGroup
+		var bookmarkManagerAddresses sync.Map
+		goroutineCount := 50
+		wait.Add(goroutineCount)
+		for i := 0; i < goroutineCount; i++ {
+			go func(i int) {
+				callExecuteQueryOrBookmarkManagerGetter(driver, i)
```
robsdedude, line 299: Might be worth to run this test twice. Once with `Driver.GetDefaultManagedBookmarkManager` being called first and once with `driver.ExecuteQuery` first.

testkit-backend/backend.go
```diff
@@ -484,6 +484,64 @@ func (b *backend) handleRequest(req map[string]any) {
 			"protocolVersion": fmt.Sprintf("%d.%d", protocolVersion.Major, protocolVersion.Minor),
 		})
 
+	case "ExecuteQuery":
+		driver := b.drivers[data["driverId"].(string)]
+		var configurers []neo4j.ExecuteQueryConfigurationOption
+		if rawConfig := data["config"]; rawConfig != nil {
+			executeQueryConfig := rawConfig.(map[string]any)
+			configurers = append(configurers, func(config *neo4j.ExecuteQueryConfiguration) {
```
robsdedude, line 502: Why is `configurers` an array but you only append once? How about having always one configurer that's just a noop (e.g., returns straight away) if `rawConfig := data["config"]` is `nil`? Might be easier to read, I'm not entirely sure.

robsdedude (approved):
A beauty! :pleading_face:

neo4j/driver_with_context.go
fbiville: yep, I've thought about it for a second and forgot :|

neo4j/driver_with_context.go
fbiville: I think for slices, it does not matter that much. I just reused the existing type of `ResultWithContext#Collect`.
In the general, passing a pointer of struct avoid copying the said struct instance.

neo4j/driver_with_context_examples_test.go
fbiville: That's specific to the Go documentation toolchain: https://go.dev/blog/examples.
Examples will appear co-located to the docs.

neo4j/driver_with_context_test.go
fbiville: go fmt is king here :)

testkit-backend/backend.go
fbiville: It's easier to turn an empty array into a vararg than `nil` (there is no ternary syntax for if-else in Go so I picked the slightly-stranger-but-less-verbose option here)

neo4j/driver_with_context_examples_test.go
robsdedude: Nice eco-system feature!

neo4j/driver_with_context.go
robsdedude: Here are some name schemes I brainstormed
* `BabysFirstBookmarkManager`
  * Together with renaming `executeQuery` to `BabysFirstQueryExecutor`
* /joking
* `ManagedBookmarkManager`
* `QueryBookmarkManager`
* `ExecuteQueryBookmarkManager`
* `Eager(Result)BookmarkManager`
* `Auto(matic)BookmarkManager`
* `DriverLevelBookmarkManager` (my current favorite, if we decide to use it by default for all driver-level execution APIs)
* `DefaultBookmarkManager` (as the ADR says, misleading: it's not being used for sessions by default)
* `SimpleApiBookmarkManager`
* `HighLevelBookmarkManager`

AndyHeap-NeoTech: (Get)BookmarkManagerSingleton
(Get)SingletonBookmarkManager
BookmarkManagerInstance
StaticBookmarkManager
GlobalBookmarkManager
BasicBookmarkManager
SuppliedBookmarkManager
BaseBookmarkManager
IncludedBookmarkManager
ProvidedBookmarkManager
AdministeredBookmarkManager

robsdedude: > `GlobalBookmarkManager`

I argue this is also misleading as it's not the same BMM across multiple drivers.

> `StaticBookmarkManager`

I like that one even though it's not really static in the Java sense.

So far I don't find any one to stand out. There are just some that I def. don't like.

neo4j/driver_with_context_examples_test.go
fbiville: let's tackle this here: https://github.com/neo4j/neo4j-go-driver/pull/424

neo4j/driver_with_context_examples_test.go
fbiville: fixed in 5020b8e7b66285f8b3fb62f67c45b7b7c4043ad8

neo4j/driver_with_context.go
fbiville: Still unsure what a good prefix or suffix would be for this...

neo4j/driver_with_context.go
fbiville: fixed in 354f54e536a7be532e5f65c5819061ae90f9409a

neo4j/driver_with_context.go
fbiville: Picked `DefaultExecuteQueryBookmarkManager` for now (see c03dea4c04433046f66b5f600bc89b730b93e964)

bigmontz (approved):

Issue #373, opened 2022-06-09
Support generics


rlch:
Hey there! Thanks for the great driver, works great.

Any plans to support generics? It would be awesome if, for instance, `ReadTransaction` had something like the following signature:
```go
ReadTransaction[T any](work TransactionWork[T], configurers ...func(*TransactionConfig)) (T, error)
```
Currently I've built a wrapper around `Session` to support my use-case, but it would be great to see this implemented within the package.

fbiville:
Hello @rich, apologies for the delay in answering, I missed the notification, it seems.

The driver will be fully able to embrace generics, when generic functions on interfaces are supported. 
This is not yet the case in 1.18.
So cases like `ReadTransaction` (also known as `ExecuteRead`) are stuck with a pre-generics signature, since it lives on the `Session` interface.

Until then, the best the 5.0 driver can offer is here: https://github.com/neo4j/neo4j-go-driver/blob/5.0/neo4j/result_helpers.go. This exposes a few standalone generic functions to extract results.

I'm of course open to ideas for improvements.

fbiville:
To add to my initial answer, let me elaborate on why the Go driver cannot fully embrace generics (let's take `Session.Run`* as an example).

A generic version of `Session` that would work today (Go 1.19) would be like this:

```go
type Session[T any] interface {
    // [...]
    Run(ctx context.Context, cypher string, params map[string]any, configurers ...func(*TransactionConfig)) (Result[T], error)
}

// assuming SessionConfig would include:

type SessionConfig[T any] interface {
    // [...]
    ResultMapper func(Result) T
}
```

The biggest issue I have with this is that a session instance is tied to a single type constraint definition.

Users wouldn't be able to call `Run` for a query that returns a `string` and another one that returns a `int64` unless they define a type constraint of `string | int64`.

You can see how this does not scale well with the number of calls one has to do on `Session`. 

Moreover, a type assertion would still be necessary to get the actual type required for each particular call, so the type assertion ceremony would still be needed.

The situation is actually worse than that, since the type constraint need to be propagated back to the session creation API, which is defined at the driver level.

As we've seen before, type constraints in interfaces cannot be declared at the function level, they need to be at the type level, so we would end up with:

```go
type Driver[T any] interface {
    // [...]
    NewSession(config SessionConfig[T]) Session[T]
}
```

It is strongly advised that `Driver` is instantiated once for the whole application lifetime. This would mean that `T` needs to be a union of all possible result types.

[...]

PR #416, opened 2022-12-13
Add generic GetProperty on nodes and relationships

fbiville:
These are generic helpers to extract properties from nodes and relationships in a type-safe way.

Relates to #373

### DIFF
--- a/neo4j/dbtype/graph.go
+++ b/neo4j/dbtype/graph.go
@@ -20,6 +20,13 @@
 // Package dbtype contains definitions of supported database types.
 package dbtype
 
+type Entity interface {
+	// Deprecated: GetId is deprecated and will be removed in 6.0. Use GetElementId instead.
+	GetId() int64
+	GetElementId() string
+	GetProperties() map[string]any
+}
+
 // Node represents a node in the neo4j graph database
 type Node struct {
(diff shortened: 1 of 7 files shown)


AndyHeap-NeoTech (approved):
Useful additions!

neo4j/dbtype/graph.go
```diff
@@ -44,6 +63,18 @@ type Relationship struct {
...
+	return r.Id
+}
+
+func (r Relationship) GetElementId() string {
+	return r.ElementId
+}
+
+func (r Relationship) GetProperties() map[string]any {
+	return r.Props
+}
```
bigmontz, line 76: Are this changes related with add generic to GetProperty?

bigmontz (approved):
⚙️

AndyHeap-NeoTech (approved):
Nice refinement to the original PR.

neo4j/dbtype/graph.go
fbiville: `GetProperties` is called in `GetProperty[T]` as follows:
```
rawValue, found := entity.GetProperties()[key]
```
I added the rest since Id/ElementId is kinda what constitutes the minimum definition of a Neo4j entity but it's not strictly needed.

bigmontz (approved):
🧂

PR #424, opened 2023-01-13
Clarify ResultWithContext#Single API docs

fbiville:
In particular, highlight that Single consumes the result cursor.

### DIFF
--- a/neo4j/result_with_context.go
+++ b/neo4j/result_with_context.go
@@ -46,7 +46,7 @@ type ResultWithContext interface {
 	// Collect fetches all remaining records and returns them.
 	Collect(ctx context.Context) ([]*Record, error)
 	// Single returns the only remaining record from the stream.
-	// If none or more than one record are left, an error is returned.
+	// If none or more than one record is left, an error is returned.
 	// The result is fully consumed after this call and its summary is immediately available when calling Consume.
 	Single(ctx context.Context) (*Record, error)
 	// Consume discards all remaining records and returns the summary information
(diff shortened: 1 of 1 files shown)


neo4j/result_with_context.go
```diff
@@ -45,8 +45,9 @@ type ResultWithContext interface {
 	Record() *Record
 	// Collect fetches all remaining records and returns them.
 	Collect(ctx context.Context) ([]*Record, error)
-	// Single returns one and only one record from the stream.
-	// If the result stream contains zero or more than one records, error is returned.
+	// Single returns the only remaining record from the stream.
+	// If none or more than one record are left, an error is returned.
```
robsdedude:
```suggestion
	// If none or more than one record is left, an error is returned.
```

neo4j/result_with_context.go
```diff
@@ -45,8 +45,9 @@ type ResultWithContext interface {
...
+	// The result is fully consumed after this call and its summary is immediately available when calling Consume.
```
robsdedude, line 50: Could `Consume` be a link? https://go.dev/doc/comment#doclinks

robsdedude (approved):

neo4j/result_with_context.go
fbiville: Turns out pluralizing the verb was both wrong in English AND French 😅

neo4j/result_with_context.go
fbiville: So I tried your suggestion with `godoc` and it does not seem to link anywhere. Even bumping the project to Go 1.19 does not seem to do the trick.
Goland autolinks both the current and suggested version.

fbiville: Let's skip the autolinking for now. If I manage to get it work, I'll have to update more than this API anyway.

PR #430, opened 2023-01-19
Clarify context.Context impact on relevant APIs

fbiville:


neo4j/session_with_context.go
```diff
@@ -64,6 +64,10 @@ type SessionWithContext interface {
 	// Run executes an auto-commit statement and returns a result
 	Run(ctx context.Context, cypher string, params map[string]any, configurers ...func(*TransactionConfig)) (ResultWithContext, error)
 	// Close closes any open resources and marks this session as unusable
+	// While Close accepts any context, it is strongly advised to avoid using contexts with timeouts or deadlines.
+	// Better use context.Background() or context.WithValue(context.Context,any).
+	// Indeed, this is the earliest chance to tell the server to clean up interrupted connections.
+	// As a consequence, this must not be interrupted or the server will accumulate connections and run out of resources.
```
robsdedude: I'm not sure this is enough. What if some workload got stuck in the middle of writing data to the socket. E.g., say half a `RUN` message has been written, but then a timeout kicked it. What will happen when the user calls `session.Close(context.Background())`. I don't know the control flow of the Go driver in-depth enough to judge if this is or isn't a problem, but I assume it could be.

fbiville: Contrary to what I thought, closing the session is currently not sufficient, since the context termination is considered a fatal error, which means the pool will unregister the terminated connection without touching it.
Because the pool (or any other component) won't close the connection, the server will hold on transactions until the connection closes.
The connection would only close when the driver (and its entire  connection pool) closes.

To avoid these issues, https://github.com/neo4j/neo4j-go-driver/pull/432 makes it so that the Bolt connection is not only marked as dead but its underlying connection is closed when the context terminates.

fbiville: It's still a good idea to not overconstrain `session.Close`since it needs time to send GOODBYE messages for each of the old-enough idle connections and clear obsolete routing table entries.
Same idea for `driver.Close` which needs time to close all the connections of the connection pool.

robsdedude: I see. Makes sense then. The other drivers do the pool clean-up when connection acquisition not when closing the session so it happens at the beginning. TIL the Go driver does it at the end.

robsdedude (approved):
:scroll:

neo4j/session_with_context.go
fbiville: Let me rephrase the docs to emphasize on connection cleanup, not connection interruptions.

fbiville: I forgot that the pool actually cleans itself up upon connection acquisition as well.

fbiville:
Let's merge https://github.com/neo4j/neo4j-go-driver/pull/432 first and clarify the docs afterwards.

robsdedude (approved):
:crying_cat_face:

bigmontz (approved):
🍄

PR #432, opened 2023-01-20
Close underlying connection when context has terminated

fbiville:
This allows the Bolt server to ditch the transaction closely after the context terminates.

Before this, the only way to have the Bolt server release transactions was to close the entire connection pool (which is done when closing the driver).


robsdedude (approved):
:boot::boom:

robsdedude (approved):

PR #436, opened 2023-01-26
Change default TLS min version to 1.2

fbiville:
The current default (before this commit) depends on the Go version.

This driver has a Go 1.16 baseline, meaning:

 - 1.0 is the platform default min TLS version for Go 1.16 and Go 1.17
 - 1.2 for clients in Go 1.18 and Go 1.19

This commit now hardcodes the minimum TLS version to 1.12, as a better security default.

If users do not specify the minimum TLS version, 1.2 will be set.

### DIFF
--- a/neo4j/internal/connector/connector.go
+++ b/neo4j/internal/connector/connector.go
@@ -84,7 +84,12 @@ func (c Connector) Connect(address string, boltLogger log.BoltLogger) (db.Connec
 		conn.Close()
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


fbiville:
Note: users really should comply to the [Go release policy](https://go.dev/doc/devel/release#policy) and run only one of the two latest major versions (as of today: Go 1.18 and Go 1.19).

bigmontz (approved):
😭

PR #448, opened 2023-02-09
Minimal Bolt 5.1 changes

thelonelyvulpes:
Add bolt 5.1
- Logon
- Logoff
- EDIT by @fbiville: ~Implement Hello & Logon pipeline.~


neo4j/internal/bolt/bolt5.go
```diff
@@ -241,31 +241,47 @@ func (b *bolt5) Connect(ctx context.Context, minor int, auth map[string]any, use
...
 	// Send hello message and wait for confirmation
 	b.out.appendHello(hello)
+
+	if minor > 0 {
+		b.out.appendLogon(auth)
+	}
+
 	b.out.send(ctx, b.conn)
-	succ := b.receiveSuccess(ctx)
+	helloSucc := b.receiveSuccess(ctx)
```
fbiville: you can go ahead and use long names `helloSuccess` by the way ^^

neo4j/internal/bolt/bolt5_test.go
```diff
@@ -235,31 +376,28 @@ func TestBolt5(outer *testing.T) {
 		}
 	})
 
-	outer.Run("Run auto-commit", func(t *testing.T) {
```
fbiville, line 238: has this test been entirely removed?

neo4j/internal/bolt/bolt5.go
```diff
@@ -274,6 +290,27 @@ func (b *bolt5) Connect(ctx context.Context, minor int, auth map[string]any, use
 	return nil
 }
 
+func (b *bolt5) ChangeUser(ctx context.Context, auth map[string]any) error {
```
fbiville: I think it makes sense to export that function, but it may be too early yet, since it's not used anywhere else?
Once we need to expose that at the driver/session level, `ChangeUser` will probably need to be defined at the `Connection` interface level as well

neo4j/internal/bolt/bolt5_test.go
thelonelyvulpes: I think I accidentally cut instead of copied.

fbiville:
Let's scope this down a bit, since the logon-logoff sequence will have to be handled differently.

AndyHeap-NeoTech (approved):
All looks present and correct.

robsdedude (approved):
Looks fine. I'd like to see tests added that the credentials never end up in the logs. I'm aware that those weren't in place before, but they really should.

Issue #451, opened 2023-02-15
Getting SIGSEGV: segmentation violation panic when using session.Run after Neo recovery


jenyad20:
[test.log](https://github.com/neo4j/neo4j-go-driver/files/10742090/test.log)
**Neo4j Version:** 3.5.23 Community  
**Neo4j Mode**: HA cluster / bolt
**Driver version**: Go driver 5.5.0, also happened on version 3  
**Operating System:** Ubuntu 22 locally, ubuntu 14 on aws

We noticed that every time a neo instance restarts our service get a panic, while investigating i found out that the panic doesn't happen when the neo is down, but rather when it recovers. I've managed to reproduce this by running the service locally,  which access neo, bombarding it with requests to reproduce production behaviour and in the middle blocking access to neo, and then restoring it.

This only happens under high load when there are requests from multiple go routines after the connection was lost and is being reestablished, 

To simulate this behaviour in testing I've created a test that opens a 100 go routines that close connection to neo and then try to execute a query, we have an error wrapper for closed connection that we reestablish in such case, and if session.Run would return an error we could handle it, however it throws a panic due to some race condition with thread unsafe buffer usage, the only solution for this would be to add a mutex that would let one go routine at a time run, but it would heavily slow down our process.

The test function
```
func (s *DriverTestSuite) TestMultithreadedQueryRequestsWithConnectionRecovery() {
	s.driver.Close(s.ctx)
	count := 100
	wg := &sync.WaitGroup{}
	wg.Add(count)

	for i := 0; i < count; i++ {

		go func(wg *sync.WaitGroup, i int, s *DriverTestSuite) {
			defer wg.Done()
			s.driver.Close(s.ctx)
			err := s.executeSimpleQuery()
			s.Require().NoError(err)

		}(wg, i, s)
	}
	wg.Wait()

}

func executeSimpleQuery(ctx context.Context, driver *Driver) error {
	return driver.ExecuteQuery(ctx, "CREATE (test:Test) return true", map[string]interface{}{}, func(result neo4j.ResultWithContext) error {
		var record *neo4j.Record
		result.NextRecord(ctx, &record)
		if record == nil || len(record.Values) == 0 {
			return errors.New("no records")
		}
		_, ok := record.Values[0].(bool)
		if !ok {
			return errors.New("expected value to be bool")
		}
		return nil
	})
}


func (s *DriverTestSuite) executeSimpleQuery() error {
[...]

PR #452, opened 2023-02-16
Try setting up a reproducer for #451

fbiville:


fbiville:
Experiment moved to https://github.com/fbiville/neo4j-go-driver-issue-451

PR #453, opened 2023-02-23
Implement user switching

fbiville:


neo4j/internal/bolt/bolt5.go
```diff
@@ -290,23 +293,21 @@ func (b *bolt5) Connect(ctx context.Context, minor int, auth map[string]any, use
...
-	b.out.send(ctx, b.conn)
-	b.receiveSuccess(ctx)
-	if b.err != nil {
-		return b.err
+	if err := b.assertState(bolt5Ready); err != nil {
+		return err
 	}
-	b.receiveSuccess(ctx)
-	return b.err
+	if !collections.ShallowMapEquals(b.authToken, authToken) {
```
robsdedude: Are you sure that shallow comparison is sufficient? If you look at `CustomAuth` for instance, it has a nested `parameters` map.

neo4j/internal/bolt/bolt5.go
```diff
@@ -575,30 +580,25 @@ func (b *bolt5) run(ctx context.Context, cypher string, params map[string]any, f
...
 	b.out.appendRun(cypher, params, meta)
-
-	// Ensure that fetchSize is in a valid range
-	switch {
-	case fetchSize < 0:
-		fetchSize = -1
-	case fetchSize == 0:
-		fetchSize = bolt5FetchSize
+	if tx == nil { // autocommit
+		b.appendReAuth() // needs to be appended *before* PULL, since PULL response is not checked here
```
robsdedude: Why would you do the re-auth after the run? That would make the run be executed as the wrong user :scream_cat:

robsdedude: Generally, I'm confused by the responsibilities here. `run` seems to be an internal helper used for tx and auto-commit tx. tx manages re-auth explicitly, auto-commit relies on `run` to handle it :thinking: 

Ideally, I think, the `appendReAuth` call should be done by the pool right after picking up the connection. However, I do see the problem that the connections as they are right not in the Go driver are not aware of potentially outstanding responses. So they might interpret the `SUCCESS` messages that come in as a response to `LOGIN/OFF` as a response to `RUN` or similar.

But it might be worth the refactoring because the approach take here is also not very future proof. It makes the assumption that after a connection has been picked up by the pool, the first thing that happens must be either a `RUN` or a `BEGIN`. That might be true today (spoiler alert, it isn't, it could also be `ROUTE`), but when changing that in the future by adding new messages to the protocol, this could very easily be overlooked.

robsdedude (changes requested):

neo4j/internal/bolt/bolt5.go
fbiville: > Why would you do the re-auth after the run? That would make the run be executed as the wrong user 🙀

ahah, 🤦 
will fix that

fbiville: > the appendReAuth call should be done by the pool right after picking up the connection. 

I think there is value in re-auth'ing the connection in a lazy fashion, but you also me realize this is much harder to do properly (kinda forgot that `ROUTE` could have triggered the re-auth)

neo4j/internal/bolt/bolt5.go
fbiville: Reflection-based comparison, here we gooo

neo4j/internal/bolt/bolt5.go
fbiville: Not resolved per se, but identified at least. Closing for now, we'll investigate a better approach tomorrow.

PR #457, opened 2023-03-03
Introduce message queue to better correlate requests and responses

fbiville:
This is currently only applied to Bolt 5.x, but can be ported to
older protocol connections as well.

This removes some fragile heuristics on response correlation
(see e.g. success#isResetResponse).

This also paves the way for a cleaner implementation of re-auth.

Big kudos to @robsdedude for guiding through this, as he already
implemented a similar approach in the Python driver.

### DIFF
--- a/neo4j/internal/bolt/bolt5.go
+++ b/neo4j/internal/bolt/bolt5.go
@@ -951,14 +951,12 @@ func (b *bolt5) resetResponseHandler() responseHandler {
 		onSuccess: func(resetSuccess *success) {
 			b.state = bolt5Ready
 		},
-		onRecord: onRecordNoOp,
 		onFailure: func(*db.Neo4jError) {
 			b.state = bolt5Dead
 		},
 		onUnknown: func(any) {
 			b.state = bolt5Dead
 		},
-		onIgnored: onIgnoredNoOp,
 	}
 }
 
--- a/neo4j/internal/bolt/bolt5.go
+++ b/neo4j/internal/bolt/bolt5.go
@@ -478,7 +478,7 @@ func (b *bolt5) pauseStream(ctx context.Context) {
 // resumeStream marks the current stream as current and requests PULL
 func (b *bolt5) resumeStream(ctx context.Context, s *stream) {
 	b.streams.resume(s)
-	b.queue.appendPullN(s.fetchSize, b.pullResponseHandler(s))
+	b.appendPullN(s)
 	b.queue.send(ctx)
 }
 
--- a/neo4j/internal/bolt/bolt5.go
+++ b/neo4j/internal/bolt/bolt5.go
@@ -483,7 +483,7 @@ func (b *bolt5) sendPullN(ctx context.Context) {
 	}
 }
 
-// Collects all records in current stream if in streaming state and there is a current stream.
+// bufferStream pulls all the records of the current stream if there is a current stream.
 func (b *bolt5) bufferStream(ctx context.Context) {
 	stream := b.streams.curr
 	if stream == nil {
@@ -505,8 +505,7 @@ func (b *bolt5) bufferStream(ctx context.Context) {
 	}
 }
 
-// Prepares the current stream for being switched out by collecting all records in the current
-// stream up until the next batch. Assumes that we are in a streaming state.
+// pauseStream pulls all the records of the current stream ongoing batch of records and unsets the stream as current
 func (b *bolt5) pauseStream(ctx context.Context) {
 	stream := b.streams.curr
 	if stream == nil {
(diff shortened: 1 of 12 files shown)


neo4j/internal/bolt/message_queue.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (q *messageQueue) appendHello(protocolVersion boltVersion, hello, token map[string]any,
+	helloHandler responseHandler, logonHandler responseHandler) {
+	q.protocolVersion = protocolVersion
+	q.out.appendHello(hello)
+	q.enqueueCallback(helloHandler)
+	if q.protocolVersion.greaterThanOrEqual(5, 1) {
+		q.out.appendLogon(token)
+		q.enqueueCallback(logonHandler)
+	}
```
robsdedude: I think it would make for a cleaner separation of concerns if the queue was more (ideally fully) protocol version agnostic.

In this particular case there could be an `appendHello` method on the queue and `bolt.Connect` is responsible for deciding whether logon needs to be enqueued or not.

neo4j/internal/bolt/bolt5.go
```diff
@@ -427,29 +386,31 @@ func (b *bolt5) discardStream(ctx context.Context) {
...
+		if err := b.queue.receiveAll(ctx); err != nil {
+			return
+		}
+		if b.err != nil {
+			return
+		}
+		if stream.sum != nil || stream.err != nil {
+			return
+		}
+		if stream.endOfBatch && discarded {
```
robsdedude, line 396: I think this check is not sufficient. Try writing a test to verify. This is the scenario I'm thinking of:
 1. Make a record stream with 4 records
 2. set fetch size to 2
 3. call `bolt.Next` 3 times (add another test with calling it 4 times)
 4. call `bolt.Discard`.
 
 The problem I suspect: there is nothing left to be discarded; everything has been `PULL`ed already, but the driver will enqueue a `DISCARD` anyway.

neo4j/internal/bolt/message_queue.go
```diff
@@ -0,0 +1,206 @@
...
+	q.out.appendRoute(routingContext, bookmarks, extras)
+	q.enqueueCallback(handler)
+}
+
+func (q *messageQueue) appendBegin(meta map[string]any, handler responseHandler) {
+	q.out.appendBegin(meta)
+	q.enqueueCallback(handler)
+}
+
+func (q *messageQueue) appendRun(cypher string, params, meta map[string]any, fetchSize int,
```
robsdedude: Similar to my comment on `appendHello`, I think `appendRun` should not be responsible for pipelining `PULL`, but bolt should probably do this explicitly.

neo4j/test-integration/dbconn_test.go
```diff
@@ -22,6 +22,8 @@ package test_integration
 import (
 	"context"
 	"crypto/rand"
+	"fmt"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/db"
```
robsdedude, line 26: Why did the driver dependency move up here? :detective:

neo4j/test-integration/dbconn_test.go
```diff
@@ -101,382 +102,390 @@ func TestConnectionConformance(outer *testing.T) {
...
+			//		tx, _ := c.TxBegin(context.Background(),
+			//			idb.TxConfig{Mode: idb.WriteMode})
+			//		_, err := c.RunTx(context.Background(), tx,
+			//			idb.Command{Cypher: "UNWIND RANGE (0, 1000000) AS n RETURN n"})
+			//		AssertNoError(t, err)
+			//		_, err = c.RunTx(context.Background(), tx,
+			//			idb.Command{Cypher: "UNWIND RANGE (0, 1000000) AS n RETURN n"})
+			//		AssertNoError(t, err)
+			//	},
+			//},
```
robsdedude: :eyes:

neo4j/internal/bolt/message_queue.go
fbiville: I agree and had the same thought yesterday evening!
Keeping protocol version-based logic here would make things too complicated (esp. for bolt4 and ROUTE)

neo4j/test-integration/dbconn_test.go
fbiville: something something leftover something

fbiville: I don't think this file changes very often, so maybe `go fmt` behavior changed a bit since last time?

neo4j/internal/bolt/message_queue.go
fbiville: Fixed in 294952d4d188f3e2c4b90be8158ea02414e676b4

neo4j/internal/bolt/message_queue.go
fbiville: Fixed in 50fddd2123cf63c1a23fe58bec26223b1ae84f68

neo4j/internal/bolt/bolt5.go
fbiville: if everything is pulled, then `stream.sum` should not be nil and will return earlier (see condition just above)?

fbiville: proven in 6ad32ecbd7d9d75434a1fc70e2cef1efbdff5cf6

robsdedude: Ah yes! I failed to realize the interplay with the `receiveAll` and `if stream.sum != nil || stream.err != nil` above. :ok_hand:

robsdedude (approved):
:up:

PR #458, opened 2023-03-09
Use message queue in Bolt 4 connections

fbiville:


neo4j/internal/bolt/bolt4.go
```diff
@@ -433,29 +397,31 @@ func (b *bolt4) discardStream(ctx context.Context) {
...
+			return
+		}
+		if stream.endOfBatch && discarded {
+			b.streams.remove(stream)
+			b.checkStreams()
+			return
+		}
+		discarded = true
+		stream.fetchSize = -1 // request infinite batch to consume the rest
+		if b.state == bolt5StreamingTx && stream.qid != b.lastQid {
```
robsdedude, line 419:
```suggestion
		if b.state == bolt4_streamingtx && stream.qid != b.lastQid {
```

neo4j/internal/bolt/message_queue_test.go
```diff
@@ -76,6 +76,7 @@ func TestMessageQueue(outer *testing.T) {
 				description: "ROUTE",
 				action: func(queue *messageQueue) {
 					queue.appendRoute(nil, nil, nil, handlers[msgRoute])
+					// note: appendRouteV43 is also ROUTE, so skipping it
```
robsdedude, line 79: Why does it exists then if it's exactly the same?

fbiville: it's not the same, but it uses the same message marker.

in order to keep the test suite simple and decoupled from the packstream serialization logic, it only checks the corresponding message marker is included in the accumulated bytes
as a result, none of the test cases must share the same message byte else the assertions will be moot

PR #459, opened 2023-03-10
Fix tfirst tracking for interleaved runs of same transaction

fbiville:
Since the connection instance tracks the last seen RUN's SUCCESS'
tfirst, a subsequent run of the same transaction overwrites the
tfirst of partially consumed results of a previous run.

Having each stream track their own tfirst avoids this problem.

Note: this only happens to Bolt4 and Bolt5, since it is not
possible to pull partial results with Bolt3 (only PULALL is
available).


robsdedude (approved):
:1st_place_medal:

robsdedude (approved):
:2nd_place_medal:

PR #463, opened 2023-03-14
Explicitly check for nil driver in `ExecuteQuery`

fbiville:
... for the naughty developers out there that ignore errors!

### DIFF
--- a/neo4j/driver_with_context.go
+++ b/neo4j/driver_with_context.go
@@ -452,6 +452,10 @@ func ExecuteQuery[T any](
 	newResultTransformer func() ResultTransformer[T],
 	settings ...ExecuteQueryConfigurationOption) (res T, err error) {
 
+	if driver == nil {
+		return *new(T), errors.New("driver cannot be nil")
+	}
+
 	if newResultTransformer == nil {
 		return *new(T), errors.New("nil is not a valid ResultTransformer function argument. " +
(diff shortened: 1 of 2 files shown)


neo4j/driver_with_context.go
```diff
@@ -452,6 +452,10 @@ func ExecuteQuery[T any](
 	newResultTransformer func() ResultTransformer[T],
 	settings ...ExecuteQueryConfigurationOption) (res T, err error) {
 
+	if driver == nil {
+		return *new(T), errors.New("driver cannot be nil")
```
bigmontz: is there any specific error type for argument errors? 


And why not return `null, error` instead of `new T, error`?

bigmontz (approved):
🏑

neo4j/driver_with_context.go
robsdedude: If `T` is not a pointer type, you cannot return `nil`.

robsdedude (approved):
:mag_right:

neo4j/driver_with_context.go
fbiville: > is there any specific error type for argument errors?

the driver actually defines a `UsageError` so that would be a good fit here.
As for`nil`, @robsdedude nailed it.

fbiville: Fixed in 25ab716703795f2888867bb719b8d380e3b8a408

PR #469, opened 2023-03-16
Close socket if Bolt handshake fails

fbiville:


bigmontz (approved):
🔋 

(it might be nice to have some tests :D)

PR #475, opened 2023-04-06
Expect nil values for access mode when opening session

thelonelyvulpes:
Add support for nil access mode to support https://github.com/neo4j-drivers/testkit/pull/548


robsdedude (approved):
:printer:

fbiville (approved):

PR #478, opened 2023-04-13
Support 5.dev as version

fbiville:


robsdedude (approved):
Tested locally (same fix, slightly different code structure). Works :+1:

PR #491, opened 2023-05-17
Make sure transactions never returns connections to the pool twice

### LINKED ISSUE #487: SessionWithContext.BeginTransaction on a new session fails with error "invalid state 3, expected: [0]"
Calling `SessionWithContext.BeginTransaction` on a newly created session returns this error:
```
invalid state 3, expected: [0]
```

This is the assertion that fails:
https://github.com/neo4j/neo4j-go-driver/blob/0f1f1e867c614b688394d50e8a9dff16c25f0569/neo4j/internal/bolt/bolt4.go#L303-L305

This session seems to be reused from the pool after previously being used to execute explicit transactions, and to have not been reset property when returned into the pool.
The implementation of `bolt4.TxBegin` seems to deal specifically with auto-commit queries, but not with explicit transactions:
https://github.com/neo4j/neo4j-go-driver/blob/0f1f1e867c614b688394d50e8a9dff16c25f0569/neo4j/internal/bolt/bolt4.go#L294-L301
I think this logic should be extended to deal with explicit transactions as well? Maybe replace:
```
	if b.state == bolt4_streaming {
```
with:
```
	if b.state == bolt4_streaming || b.state == bolt4_streamingtx {
```
???

I was previously using `v5.0.0-preview` where this bug didn't occur. This is a regression.

**Neo4j Version:** 4.4.15-enterprise
**Neo4j Mode**: Single instance  
**Driver version**: Go driver 5.8.x
**Operating System:** Ubuntu 16.10 on AWS

fbiville:
Before this fix, users could run into situation when a transaction
would execute their `onClosed` callback more than once, meaning
the referenced connection could be returned to the idle list of
its server's connection several times (i.e. the connection ref
could be duplicated and borrowed by more than 1 transaction at a
time, which breaks the fundamental invariant of connections).

This could occur when a transaction was committed or
rolled back* and then, users would run a query with the same transaction
again. This is obviously a misuse of the transaction API but the driver
could fail in a more graceful and more understandable manner.

When that subsequent run executed, a transaction ID mismatch error
would then be returned (since the connection was reset upon the
return to pool after commit/rollback, thus forgetting about its former
transaction ID).
The tx ID mismatch error would then lead to `onClosed` being called
a second time, and the same connection would end up listed a second time
in its server's tracked idle connection list.

One fix could be to make the pool return idempotent.
However, we felt that the better fix is to respect the invariant 
according to which a connection is never returned more than once, thus
making the pool return idempotency irrelevant in practice. 

*rollback also happens when closing the transaction or closing the 
enclosing session and the transaction is the currently active workload
in that session

Fixes #487

### DIFF
--- a/neo4j/session_with_context.go
+++ b/neo4j/session_with_context.go
@@ -308,7 +308,6 @@ func (s *sessionWithContext) BeginTransaction(ctx context.Context, configurers .
 		return nil, errorutil.WrapError(err)
 	}
 
-	// Begin transaction
 	beginBookmarks, err := s.getBookmarks(ctx)
 	if err != nil {
 		_ = s.pool.Return(ctx, conn)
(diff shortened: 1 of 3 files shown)


robsdedude (approved):
:electric_plug:

PR #494, opened 2023-05-22
Gracefully return error when no callback defined for response

### LINKED ISSUE #492: resetResponseHandler does not set an onIgnored handler
The resetResponseHandler does not set an onIgnored callback. If that handler is called for a bolt message of type "ignored" then you get a SIGSEV as follows:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5ec720]


goroutine 1 [running]:
github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt.(*messageQueue).receive(0xc0002020f8, {0x71e750, 0xc0000b6000})
        /home/trule/go/pkg/mod/github.com/neo4j/neo4j-go-driver/v5@v5.8.1/neo4j/internal/bolt/message_queue.go:166 +0x200
github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt.(*bolt4).ForceReset(0xc000202000, {0x71e750, 0xc0000b6000})
        /home/trule/go/pkg/mod/github.com/neo4j/neo4j-go-driver/v5@v5.8.1/neo4j/internal/bolt/bolt4.go:821 +0x1bc
github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt.(*bolt4).Reset(0xc000000140?, {0x71e750?, 0xc0000b6000?})
        /home/trule/go/pkg/mod/github.com/neo4j/neo4j-go-driver/v5@v5.8.1/neo4j/internal/bolt/bolt4.go:803 +0x7d
github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/pool.(*Pool).Return(0xc000000140, {0x71e750, 0xc0000b6000}, {0x720e70?, 0xc000202000?})
        /home/trule/go/pkg/mod/github.com/neo4j/neo4j-go-driver/v5@v5.8.1/neo4j/internal/pool/pool.go:450 +0x2b2
github.com/neo4j/neo4j-go-driver/v5/neo4j.(*sessionWithContext).executeTransactionFunction.func1()
        /home/trule/go/pkg/mod/github.com/neo4j/neo4j-go-driver/v5@v5.8.1/neo4j/session_with_context.go:436 +0x36
```

The callback is thus (see `onIgnored: nil`):

```
(dlv) print callback
github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt.responseHandler {
    onSuccess: github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt.(*bolt4).resetResponseHandler.func1, 
    onRecord: nil, 
    onFailure: github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt.(*bolt4).resetResponseHandler.func2, 
    onUnknown: github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt.(*bolt4).resetResponseHandler.func3, 
    onIgnored: nil
}
```

Adding `onIgnored: onIgnoredNoOp,` to the resetResponseHandler() would likely fix the fault.

https://github.com/neo4j/neo4j-go-driver/blob/0eccfa3db544f6639151bf013cb08a3dde97fc70/neo4j/internal/bolt/bolt4.go#LL1078C1-L1091C2

The scenario behind this is fairly simple (and probably stupid) ... however, I'm new to Go ... based of the example, I basically run this function twice with the same parameters, the second time fails.

```
func addNodeFunc(ctx context.Context, labels []string, name string) neo4j.ManagedTransactionWork {
	return func(tx neo4j.ManagedTransaction) (any, error) {
		var b strings.Builder
		b.WriteString("CREATE (n:" + strings.Join(labels, ":") +" {name: $name}) ")
		b.WriteString("RETURN id(n) ")
		query := b.String()
		result, err := tx.Run(ctx, query, map[string]any{"name": name})
		if err != nil {
			return nil, err
		}
		return result.Consume(ctx)
	}
}
```

fbiville:
Fixes #492

### DIFF
--- a/neo4j/internal/bolt/message_queue.go
+++ b/neo4j/internal/bolt/message_queue.go
@@ -159,32 +159,32 @@ func (q *messageQueue) receive(ctx context.Context) error {
 	case *db.Record:
 		onRecord := handler.onRecord
 		if onRecord == nil {
-			return errors.New("this response handler does not support RECORD")
+			return errors.New("protocol violation: the server sent an unexpected RECORD response")
 		}
 		onRecord(message)
 	case *success:
 		onSuccess := handler.onSuccess
 		if onSuccess == nil {
-			return errors.New("this response handler does not support SUCCESS")
+			return errors.New("protocol violation: the server sent an unexpected SUCCESS response")
 		}
 		onSuccess(message)
 	case *db.Neo4jError:
 		onFailure := handler.onFailure
 		if onFailure == nil {
-			return errors.New("this response handler does not support FAILURE")
+			return errors.New("protocol violation: the server sent an unexpected FAILURE response")
 		}
 		onFailure(ctx, message)
 		return message
 	case *ignored:
 		onIgnored := handler.onIgnored
 		if onIgnored == nil {
-			return errors.New("this response handler does not support IGNORED")
+			return errors.New("protocol violation: the server sent an unexpected IGNORED response")
 		}
 		onIgnored(message)
 	default:
 		onUnknown := handler.onUnknown
 		if onUnknown == nil {
-			return fmt.Errorf("this response handler does not support unknown %v", message)
+			return fmt.Errorf("protocol violation: the server sent an unknown %v response", message)
 		}
 		onUnknown(message)
 	}
--- a/neo4j/internal/bolt/message_queue.go
+++ b/neo4j/internal/bolt/message_queue.go
@@ -23,6 +23,7 @@ import (
 	"container/list"
 	"context"
 	"errors"
+	"fmt"
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/db"
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/log"
 	"net"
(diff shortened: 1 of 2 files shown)


neo4j/internal/bolt/message_queue.go
robsdedude, line 1: This should solve the issue indeed. I just wonder if there could be a better error message. Understanding this error requires a lot of in-depth knowledge of the driver. Also, it's saying "*this* response handler". But from a user's perspective there is no reference to which one it is (I think). Let alone what a response handler is in the first place. What should the user do? Is this their fault? If so, what actions could they try?

How about something along the lines of
`protocol violation: the server sent an unexpected RECORD response`?
And if you want to be really explicit, add something like `this is likely a bug in the server or driver implementation` or such.

fbiville: yep, the error message was half-baked, I like the first suggestion. Let me update the PR.

robsdedude (approved):
:eye_speech_bubble:

PR #497, opened 2023-05-23
Expose single record extraction utility to eager result

fbiville:


ali-ince (approved):
looks good to me.

fbiville:
I thought about @robsdedude arguments against introducing this and the one that resonated with me the most is: `EagerResult` has everything loaded in memory and `Single` is just a simple `len(myEagerResult) == 1`, so the added value is quite minimal.

On the other hand, `Result` does not have a known size and there is value in exposing `Single` which asserts there is exactly a single result left to consume.

I'll close this PR for now.

PR #501, opened 2023-06-08
Update API docs link

stefano-ottolenghi:
The current link points to the package root, and it's not immediate (at least to me) how to actually access the API docs. The path I had to take is search for a symbol, and then strip part of the URL, and I think that is the correct link we want.


fbiville (approved):

fbiville:
Thanks!

PR #506, opened 2023-06-13
Add commented blank line to fix `Driver` deprecation notice

### LINKED ISSUE #502: `Driver` type does not show as Deprecated in API docs
`NewDriver` shows as deprecated, while `Driver` does not. 

Screenshot from 2023-06-13 04-07-37

fbiville:
Fixes #502

### DIFF
--- a/neo4j/result.go
+++ b/neo4j/result.go
@@ -23,7 +23,7 @@ import (
 	"context"
 )
 
-// Result is created via the context-aware Session type.
+// Result is created via the Session type.
 //
 // Deprecated: use ResultWithContext instead.
 type Result interface {
--- a/neo4j/result.go
+++ b/neo4j/result.go
@@ -23,8 +23,9 @@ import (
 	"context"
 )
 
+// Result is created via the context-aware Session type.
+//
 // Deprecated: use ResultWithContext instead.
-// ResultWithContext is created via the context-aware SessionWithContext type.
 type Result interface {
 	// Keys returns the keys available on the result set.
 	Keys() ([]string, error)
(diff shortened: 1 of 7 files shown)


bigmontz (approved):
📝

stefano-ottolenghi (approved):

neo4j/resultsummary.go
```diff
@@ -218,8 +218,9 @@ type Notification interface {
 	// Not all notifications have a unique position to point to and in that case the position would be set to nil.
 	Position() InputPosition
 	// Severity returns the severity level of this notification.
-	// Deprecated: Severity will be removed in 6.0.
 	// Please rely on SeverityLevel (or RawSeverityLevel) instead.
+	//
+	// Deprecated: Severity will be removed in 6.0.
```
robsdedude: That's the wrong way around, isn't it? Shouldn't the "pleas use XYZ" be part of the deprecation message?

fbiville: fixed in f6df686a5c69a0375f26b2521d7ff2e18bae67b2

neo4j/result.go
```diff
@@ -23,8 +23,9 @@ import (
 	"context"
 )
 
+// Result is created via the context-aware Session type.
```
robsdedude: Is it? I thought, at least as it stands right now, it's created via either a session (with or without context) or a transaction (with or without context).

robsdedude (approved):

neo4j/result.go
fbiville: yeah, my fix is bogus
`Result` is created via `Session`, which is **not** "context-aware.
`ResultWithContext` is created via `SessionWithContext`, which is "context-aware".

StephenCathcart:
```suggestion
// Result is created via the Session type.
```

StephenCathcart (changes requested):
👍

PR #510, opened 2023-07-04
Spatial types stringers to show actual field names

stefano-ottolenghi:
I find it more helpful to get the actual struct field names, saves a trip to the API docs when your query returns one of these objects (especially the case can be confusing).


StephenCathcart (commented):
LGTM!

PR #515, opened 2023-07-10
feat: support connect failed to force refresh routing table

youngxhui:
when one of the neo4j cluster's instance shutdown, the connection will be failed, becuase the routing table ttl is 300s. In the 300s, the routing table can not refresh, the session is failed until next routing table refresh. if modifiy router ttl's value short on neo4j server, the cache is no valueable .

now, if session connection failed, client will be forced refresh routing table and retry connection neo4j server. 

```go
s.router.Invalidate(s.databaseName)
```

The router make the database invalidate, when it next try to connection neo4j cluster ,the router will be create new routing table.

### DIFF
--- a/neo4j/session.go
+++ b/neo4j/session.go
@@ -282,6 +282,9 @@ func (s *session) runRetriable(
 	for state.Continue() {
 		if workResult, successfullyCompleted := s.tryRun(&state, mode, &config, work); successfullyCompleted {
 			return workResult, nil
+		} else {
+			// Force refresh table routing
+			s.router.Invalidate(s.databaseName)
 		}
 	}
 
--- a/neo4j/session.go
+++ b/neo4j/session.go
@@ -283,7 +283,7 @@ func (s *session) runRetriable(
 		if workResult, successfullyCompleted := s.tryRun(&state, mode, &config, work); successfullyCompleted {
 			return workResult, nil
 		} else {
-			// Force refresh table routing
+			// Force refresh routing table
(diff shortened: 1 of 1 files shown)


StephenCathcart:
As per issue #519, a fix for the problem mentioned will be included in a future release.

robsdedude:
Thank you very much for taking the time to make a contribution @youngxhui :raised_hands: 

As mentioned by @StephenCathcart we've been working on a fix for this issue independently. 
https://github.com/neo4j/neo4j-go-driver/pull/520 should fix it while even being broader: removing bad servers from the routing table should not only happen when using transaction functions (`session.runRetriable`), but for any other API as well.

I will close this PR for now. However, we will come back to it should we not move forward with the other PR I've linked.

youngxhui:
> Thank you very much for taking the time to make a contribution @youngxhui 🙌
> 
> As mentioned by @StephenCathcart we've been working on a fix for this issue independently. #520 should fix it while even being broader: removing bad servers from the routing table should not only happen when using transaction functions (`session.runRetriable`), but for any other API as well.
> 
> I will close this PR for now. However, we will come back to it should we not move forward with the other PR I've linked.

Is there a plan to support 4.4?

StephenCathcart:
Hi @youngxhui, unfortunately for now there are no plans to support this change in 4.4 - there are too many things to backport for an LTS. Upgrading to the 5.x driver is _fairly_ straightforward, if you don't mind deprecation warnings.

PR #516, opened 2023-07-18
Handle sub-millisecond transaction timeout

StephenCathcart:


neo4j/internal/bolt/bolt3_test.go
```diff
@@ -934,4 +934,69 @@ func TestBolt3(outer *testing.T) {
...
+		input       time.Duration
+		output      int
+		omitted     bool
+	}
+
+	txTimeoutTestCases := []txTimeoutTestCase{
+		{
+			description: "tx_timeout should not be present",
+			input:       0,
+			output:      0,
```
robsdedude:
```suggestion
```
It doesn't matter what the output is, as we don't test it. So I don't see the benefit of being explicit here. Also, `0` is the default value anyway.

same comment for the other test files (won't repeat).

robsdedude (approved):
Optionally add tests for the log message.

PR #520, opened 2023-08-08
Catching up with TestKit

robsdedude:
1. Add support for routing table specific TestKit messaged (retrieving routing tables and forcing updates)
2. Improve logging in the TestKit backend (all logs-driver and bolt-got both to stdout and TestKit) for easier debugging
3. Overhaul how/when the driver drops servers from the cached routing table:
   * Move logic of deactivating servers on failures into the pool. This functionality should not only be present when using transaction functions.
   * Drop writers on certain error codes: `Neo.ClientError.Cluster.NotALeader` and `Neo.ClientError.General.ForbiddenOnReadOnlyDatabase`
4. To simplify the code and avoid deadlocks or inconsistent driver states, both pool and router use blocking locks now. However, they will never perform IO while holding the lock removing the need for lock acquisition timeouts.

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/587


neo4j/driver_with_context.go
```diff
@@ -394,9 +397,7 @@ func (d *driverWithContext) Close(ctx context.Context) error {
 	defer d.mut.Unlock()
 	// Safeguard against closing more than once
 	if d.pool != nil {
-		if err := d.pool.Close(ctx); err != nil {
-			return err
```
StephenCathcart, line 398: Is there a reason we want to remove the error check of pool.Close()?

StephenCathcart: Ignore my comment - func has removed error return 👍

StephenCathcart (approved):
LGTM! :shipit:

neo4j/internal/bolt/bolt5.go
```diff
@@ -921,19 +922,23 @@ func (b *bolt5) reAuth(ctx context.Context, auth *idb.ReAuthToken) error {
...
 		b.log.Warnf(log.Driver, b.serverName, "could not close underlying socket")
 	}
-	b.state = bolt5Dead
 }
 
 func (b *bolt5) SelectDatabase(database string) {
 	b.databaseName = database
 }
 
+func (b *bolt5) SelectedDatabase() string {
```
fbiville: nit: I'd personally go for just `Database()`

neo4j/internal/bolt/connections.go
```diff
@@ -27,7 +27,11 @@ import (
 	"net"
 )
 
-type Neo4jErrorCallback func(context.Context, idb.Connection, *db.Neo4jError) error
+type ConnectionErrorListener interface {
+	OnNeo4jError(context.Context, idb.Connection, *db.Neo4jError) error
```
fbiville, line 31: why do they return errors?

testkit-backend/backend.go
```diff
@@ -150,6 +150,7 @@ func (b *backend) writeLine(s string) error {
 func (b *backend) writeLineLocked(s string) error {
 	b.wrLock.Lock()
 	defer b.wrLock.Unlock()
+	fmt.Println(s)
```
fbiville, line 153: I think this was added on purpose, but I forgot why. Can you tell (again)?

fbiville (approved):

testkit-backend/backend.go
robsdedude: So that log output goes to both sides: driver (this `Println`) and TestKit (`return b.writeLine(s)` in the next line).

neo4j/internal/bolt/connections.go
robsdedude: `pool.OnNeo4jError` might call the auth manager which is fallible.

The other methods used to be fallible because of the racing locks they had. They don't any longer so this could be changed. I'm just wondering how valid it is to assume that the pool might want to start failing in those functions in the future. Not sure. What do you think?

fbiville: Those are internal, so no need to be future-proof, I'd say. If they don't return errors, let's just remove these from the signatures?

robsdedude: :zap:

PR #521, opened 2023-08-16
Fix race conditions in unit tests

robsdedude:
Ideally, we'd also be turning on the `-race` flag for tests. But this makes the tests stuck. Needs investigation.


DavidS-ovm:
See also https://github.com/neo4j/neo4j-go-driver/issues/529

StephenCathcart (approved):
🍏

PR #522, opened 2023-08-18
Pipelined ExecuteQuery

thelonelyvulpes:
Implement pipelined `BEGIN` message and handling for `ExecuteQuery`

### DIFF
--- a/neo4j/internal/bolt/bolt4_test.go
+++ b/neo4j/internal/bolt/bolt4_test.go
@@ -502,6 +502,7 @@ func TestBolt4(outer *testing.T) {
 								DisCats: test.DisCats,
 							},
 						},
+						true,
 					)
 				}
 				if test.ExpectError {
@@ -523,7 +524,7 @@ func TestBolt4(outer *testing.T) {
 		defer bolt.Close(context.Background())
 
 		tx, err := bolt.TxBegin(context.Background(),
-			idb.TxConfig{Mode: idb.ReadMode})
+			idb.TxConfig{Mode: idb.ReadMode}, true)
 		AssertNoError(t, err)
 		assertBoltState(t, bolt4_tx, bolt)
 		str, err := bolt.RunTx(context.Background(), tx,
@@ -571,7 +572,7 @@ func TestBolt4(outer *testing.T) {
 		defer bolt.Close(context.Background())
 
 		tx, err := bolt.TxBegin(context.Background(),
-			idb.TxConfig{Mode: idb.ReadMode})
+			idb.TxConfig{Mode: idb.ReadMode}, true)
 		AssertNoError(t, err)
 		_, err = bolt.RunTx(context.Background(), tx,
 			idb.Command{Cypher: "Whatever", FetchSize: 1})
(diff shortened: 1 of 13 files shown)


neo4j/internal/bolt/bolt5.go
```diff
@@ -324,14 +325,16 @@ func (b *bolt5) TxBegin(
...
+		}
+		if err := b.queue.receiveAll(ctx); err != nil {
+			return 0, err
+		}
+		if b.err != nil { // onNextMessageErr kicked in
+			return 0, b.err
+		}
 	}
 
 	b.state = bolt5Tx
```
thelonelyvulpes: we are not actually in tx state on the server so is setting `b.state` to `bolt5Tx` an issue? likewise in bolt4

robsdedude: Not only is it not an issue, it's necessary. Further down down in `bolt5.run` you'll see
```go
	if b.state == bolt5Ready {
		b.state = bolt5Streaming
	} else if b.state == bolt5Tx {
		b.state = bolt5StreamingTx
	}
```

That would put the connection if the wrong state on the pipelined `RUN` if `begin` weren't setting `state` to `bolt5Tx`.

neo4j/internal/bolt/bolt5.go
```diff
@@ -324,14 +325,16 @@ func (b *bolt5) TxBegin(
...
+	if syncMessages {
+		if b.queue.send(ctx); b.err != nil {
+			return 0, b.err
+		}
+		if err := b.queue.receiveAll(ctx); err != nil {
+			return 0, err
+		}
+		if b.err != nil { // onNextMessageErr kicked in
+			return 0, b.err
+		}
```
robsdedude: This check needs to happen regardless if the message is sent or not. `appendBegin` might call the queue's `outgoing.onPackErr` which will set the `b.err`. You might be able to add a unit test to see that in action. Just attach unserializable metadata to the transaction.

robsdedude (changes requested):

neo4j/internal/bolt/bolt4_test.go
```diff
@@ -671,6 +672,42 @@ func TestBolt4(outer *testing.T) {
...
+		})
+		defer cleanup()
+		defer bolt.Close(context.Background())
+
+		id, err := bolt.TxBegin(context.Background(),
+			idb.TxConfig{Mode: idb.ReadMode, Bookmarks: []string{"bm1"}}, false)
+
+		assertBoltState(t, bolt4_tx, bolt)
+		AssertNoError(t, err)
+		AssertNotDeepEquals(t, id, 0)
```
robsdedude: What does this assert? Does `0` have a special semantic?

thelonelyvulpes: It could be removed as the `AssertNoError` will fail first, but all of the failure branches in the method return `0` for the id

robsdedude: That's just because it needs to return something along with the error. Any other number would work just as fine.

Try suggesting enum/union types in Go. That'd improve the situation here. ;)

robsdedude (approved):
:syringe:

StephenCathcart:
Looks good to me! 🚗

StephenCathcart (approved):

PR #523, opened 2023-08-22
ADR 019 password rotation

StephenCathcart:
This PR updates the preview feature "re-auth" significantly. The changes allow for catering to a wider range of use cases including simple password rotation.

`ExpirationBasedTokenManager` was renamed to `BearerTokenManager` for handling potentially expiring auth information:

```go
func ExampleBearerTokenManager() {
	fetchAuthTokenFromMyProvider := func(ctx context.Context) (neo4j.AuthToken, *time.Time, error) {
		// some way of getting a token
		token, err := getSsoToken(ctx)
		if err != nil {
			return neo4j.AuthToken{}, nil, err
		}
		// assume we know our tokens expire every 60 seconds
		expiresIn := time.Now().Add(60 * time.Second)
		// Include a little buffer so that we fetch a new token *before* the old one expires
		expiresIn = expiresIn.Add(-10 * time.Second)
		// or return nil instead of `&expiresIn` if we don't expect it to expire
		return token, &expiresIn, nil
	}
	// create a new driver with a bearer token manager which uses provider to handle possibly expiring auth tokens.
	_, _ = neo4j.NewDriverWithContext(getUrl(), auth.BearerTokenManager(fetchAuthTokenFromMyProvider))
}
```

A `BasicTokenManager` token manager was added to handle password rotation:
```go
func ExampleBasicTokenManager() {
	fetchBasicAuthToken := func(ctx context.Context) (neo4j.AuthToken, error) {
		// some way of getting basic authentication information
		username, password, realm, err := getBasicAuth()
		if err != nil {
			return neo4j.AuthToken{}, err
		}
		// create and return a basic authentication token with provided username, password and realm
		return neo4j.BasicAuth(username, password, realm), nil
	}
	// create a new driver with a basic token manager which uses provider to handle basic auth password rotation.
	_, _ = neo4j.NewDriverWithContext(getUrl(), auth.BasicTokenManager(fetchBasicAuthToken))
}
```

### DIFF
--- a/neo4j/auth/auth.go
+++ b/neo4j/auth/auth.go
@@ -21,7 +21,9 @@ package auth
 
 import (
 	"context"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/db"
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/auth"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/collections"
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/racing"
 	"reflect"
 	"time"
@@ -51,19 +53,23 @@ type TokenManager interface {
 	// The token returned must always belong to the same identity.
 	// Switching identities using the `TokenManager` is undefined behavior.
 	GetAuthToken(ctx context.Context) (auth.Token, error)
-	// OnTokenExpired is called by the driver when the provided token expires
-	// OnTokenExpired should invalidate the current token if it matches the provided one
-	OnTokenExpired(context.Context, auth.Token) error
+
+	// HandleSecurityException is called when the server returns any `Neo.ClientError.Security.*` error.
+	// It should return true if the error was handled, in which case the driver will mark the error as retryable.
+	HandleSecurityException(context.Context, auth.Token, *db.Neo4jError) (bool, error)
 }
+
+type authTokenProvider = func(context.Context) (auth.Token, error)
 
 type authTokenWithExpirationProvider = func(context.Context) (auth.Token, *time.Time, error)
 
 type expirationBasedTokenManager struct {
-	provider   authTokenWithExpirationProvider
-	token      *auth.Token
-	expiration *time.Time
-	mutex      racing.Mutex
-	now        *func() time.Time
+	provider             authTokenWithExpirationProvider
+	token                *auth.Token
+	expiration           *time.Time
+	mutex                racing.Mutex
+	now                  *func() time.Time
+	handledSecurityCodes collections.Set[string]
 }
 
 func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.Token, error) {
@@ -83,34 +89,76 @@ func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.To
 	return *m.token, nil
 }
 
-func (m *expirationBasedTokenManager) OnTokenExpired(ctx context.Context, token auth.Token) error {
+func (m *expirationBasedTokenManager) HandleSecurityException(ctx context.Context, token auth.Token, securityException *db.Neo4jError) (bool, error) {
+	if !m.handledSecurityCodes.Contains(securityException.Code) {
+		return false, nil
+	}
 	if !m.mutex.TryLock(ctx) {
-		return racing.LockTimeoutError(
-			"could not acquire lock in time when handling token expiration in ExpirationBasedTokenManager")
+		return false, racing.LockTimeoutError(
+			"could not acquire lock in time when handling token expiration in expirationBasedTokenManager")
 	}
 	defer m.mutex.Unlock()
 	if m.token != nil && reflect.DeepEqual(token.Tokens, m.token.Tokens) {
 		m.token = nil
 	}
-	return nil
+	return true, nil
 }
 
-// ExpirationBasedTokenManager creates a token manager for potentially expiring auth info.
-//
-// The first and only argument is a provider function that returns auth information and an optional expiration time.
-// If the expiration time is nil, the auth info is assumed to never expire.
+// Basic creates a TokenManager handling basic auth password rotation.
+// The provider function returns basic auth information and is assumed to never expire.
 //
 // WARNING:
 //
-//	The provider function *must not* interact with the driver in any way as this can cause deadlocks and undefined
-//	behaviour.
+// The provider function *must not* interact with the driver in any way as this can cause deadlocks and undefined
+// behaviour.
 //
-//	The provider function only ever return auth information belonging to the same identity.
-//	Switching identities is undefined behavior.
+// The provider function must only ever return auth information belonging to the same identity.
+// Switching identities is undefined behavior.
 //
-// ExpirationBasedTokenManager is part of the re-authentication preview feature
+// Basic is part of the re-authentication preview feature
 // (see README on what it means in terms of support and compatibility guarantees)
-func ExpirationBasedTokenManager(provider authTokenWithExpirationProvider) TokenManager {
+func Basic(provider authTokenProvider) TokenManager {
 	now := time.Now
-	return &expirationBasedTokenManager{provider: provider, mutex: racing.NewMutex(), now: &now}
+	return &expirationBasedTokenManager{
+		provider: wrapWithNilExpiration(provider),
+		mutex:    racing.NewMutex(),
+		now:      &now,
+		handledSecurityCodes: collections.NewSet([]string{
+			"Neo.ClientError.Security.Unauthorized",
+		}),
+	}
+}
+
+// Bearer creates a TokenManager handling potentially expiring auth information.
+// The provider function returns auth information and an optional expiration time.
+// If the expiration time is nil, the auth information is assumed to never expire.
+//
+// WARNING:
+//
+// The provider function *must not* interact with the driver in any way as this can cause deadlocks and undefined
+// behaviour.
+//
+// The provider function must only ever return auth information belonging to the same identity.
+// Switching identities is undefined behavior.
+//
+// Bearer is part of the re-authentication preview feature
+// (see README on what it means in terms of support and compatibility guarantees)
+func Bearer(provider authTokenWithExpirationProvider) TokenManager {
+	now := time.Now
+	return &expirationBasedTokenManager{
+		provider: provider,
+		mutex:    racing.NewMutex(),
+		now:      &now,
+		handledSecurityCodes: collections.NewSet([]string{
+			"Neo.ClientError.Security.TokenExpired",
+			"Neo.ClientError.Security.Unauthorized",
+		}),
+	}
+}
+
+func wrapWithNilExpiration(provider authTokenProvider) authTokenWithExpirationProvider {
+	return func(ctx context.Context) (auth.Token, *time.Time, error) {
+		token, err := provider(ctx)
+		return token, nil, err
+	}
 }
--- a/neo4j/auth/auth.go
+++ b/neo4j/auth/auth.go
@@ -63,7 +63,7 @@ type authTokenProvider = func(context.Context) (auth.Token, error)
 
 type authTokenWithExpirationProvider = func(context.Context) (auth.Token, *time.Time, error)
 
-type expirationBasedTokenManager struct {
+type neo4jAuthTokenManager struct {
 	provider             authTokenWithExpirationProvider
 	token                *auth.Token
 	expiration           *time.Time
@@ -72,10 +72,10 @@ type expirationBasedTokenManager struct {
 	handledSecurityCodes collections.Set[string]
 }
 
-func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.Token, error) {
+func (m *neo4jAuthTokenManager) GetAuthToken(ctx context.Context) (auth.Token, error) {
 	if !m.mutex.TryLock(ctx) {
 		return auth.Token{}, racing.LockTimeoutError(
-			"could not acquire lock in time when getting token in ExpirationBasedTokenManager")
+			"could not acquire lock in time when getting token in neo4jAuthTokenManager")
 	}
 	defer m.mutex.Unlock()
 	if m.token == nil || m.expiration != nil && (*m.now)().After(*m.expiration) {
@@ -89,13 +89,13 @@ func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.To
 	return *m.token, nil
 }
 
-func (m *expirationBasedTokenManager) HandleSecurityException(ctx context.Context, token auth.Token, securityException *db.Neo4jError) (bool, error) {
+func (m *neo4jAuthTokenManager) HandleSecurityException(ctx context.Context, token auth.Token, securityException *db.Neo4jError) (bool, error) {
 	if !m.handledSecurityCodes.Contains(securityException.Code) {
 		return false, nil
 	}
 	if !m.mutex.TryLock(ctx) {
 		return false, racing.LockTimeoutError(
-			"could not acquire lock in time when handling token expiration in expirationBasedTokenManager")
+			"could not acquire lock in time when handling security exception in neo4jAuthTokenManager")
 	}
 	defer m.mutex.Unlock()
 	if m.token != nil && reflect.DeepEqual(token.Tokens, m.token.Tokens) {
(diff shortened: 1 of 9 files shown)


StephenCathcart:
[[Go] Re-authentication on credentials changing/password rotation](https://trello.com/c/cfgvdGWt/19-go-re-authentication-on-credentials-changing-password-rotation)

neo4j/auth/auth.go
```diff
@@ -83,34 +89,76 @@ func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.To
...
-//	The provider function only ever return auth information belonging to the same identity.
-//	Switching identities is undefined behavior.
+// The provider function must only ever return auth information belonging to the same identity.
+// Switching identities is undefined behavior.
 //
-// ExpirationBasedTokenManager is part of the re-authentication preview feature
+// Bearer is part of the re-authentication preview feature
 // (see README on what it means in terms of support and compatibility guarantees)
-func ExpirationBasedTokenManager(provider authTokenWithExpirationProvider) TokenManager {
+func Bearer(provider authTokenWithExpirationProvider) TokenManager {
```
robsdedude: I'm unhappy with the naming.

`auth.Bearer(...)` reads a bit like this is a static bearer token and easily mistaken for `BearerAuth(...)`. Maybe it should be called `auth.BearerManager` or `auth.BearerTokenManager` or such.

Same applies to `Basic` above.

neo4j/auth/auth.go
```diff
@@ -51,19 +53,23 @@ type TokenManager interface {
...
+	// HandleSecurityException is called when the server returns any `Neo.ClientError.Security.*` error.
+	// It should return true if the error was handled, in which case the driver will mark the error as retryable.
+	HandleSecurityException(context.Context, auth.Token, *db.Neo4jError) (bool, error)
 }
 
+type authTokenProvider = func(context.Context) (auth.Token, error)
+
 type authTokenWithExpirationProvider = func(context.Context) (auth.Token, *time.Time, error)
 
 type expirationBasedTokenManager struct {
```
robsdedude: Some implementations renamed this thing to `neo4jAuthTokenManager` or some variant of that. It's obviously an internal name so you can call it whatever you want.

neo4j/auth/auth.go
```diff
@@ -83,34 +89,76 @@ func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.To
...
-func (m *expirationBasedTokenManager) OnTokenExpired(ctx context.Context, token auth.Token) error {
+func (m *expirationBasedTokenManager) HandleSecurityException(ctx context.Context, token auth.Token, securityException *db.Neo4jError) (bool, error) {
+	if !m.handledSecurityCodes.Contains(securityException.Code) {
+		return false, nil
+	}
 	if !m.mutex.TryLock(ctx) {
-		return racing.LockTimeoutError(
-			"could not acquire lock in time when handling token expiration in ExpirationBasedTokenManager")
+		return false, racing.LockTimeoutError(
+			"could not acquire lock in time when handling token expiration in expirationBasedTokenManager")
```
robsdedude: That message is out of date. "token expiration" is not the only thing that might be handled here.

neo4j/auth/auth.go
```diff
@@ -83,34 +89,76 @@ func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.To
...
 	if m.token != nil && reflect.DeepEqual(token.Tokens, m.token.Tokens) {
 		m.token = nil
 	}
-	return nil
+	return true, nil
 }
 
-// ExpirationBasedTokenManager creates a token manager for potentially expiring auth info.
+// Basic creates a TokenManager handling basic auth password rotation.
+// The provider function returns basic auth information and is assumed to never expire.
```
robsdedude: Might be misleading. It doesn't expire in the sense that the driver does not set an expiration time, but it's still expected to expire in the sense that the password will be changed on the server.

Stepping back and looking at the grand scheme of the docs, I wonder if this is enough to explain to a user not knowing how re-auth works what the two factory methods here do.

neo4j/auth/auth.go
```diff
@@ -83,34 +89,76 @@ func (m *expirationBasedTokenManager) GetAuthToken(ctx context.Context) (auth.To
...
 	now := time.Now
-	return &expirationBasedTokenManager{provider: provider, mutex: racing.NewMutex(), now: &now}
+	return &expirationBasedTokenManager{
+		provider: provider,
+		mutex:    racing.NewMutex(),
+		now:      &now,
+		handledSecurityCodes: collections.NewSet([]string{
+			"Neo.ClientError.Security.TokenExpired",
+			"Neo.ClientError.Security.Unauthorized",
+		}),
```
robsdedude, line 157: Ran a quick benchmark
```go
func fooSlice(container *[]string, element *string) bool {
	for _, s := range *container {
		if s == *element {
			return true
		}
	}
	return false
}

func fooSet(container *collections.Set[string], element *string) bool {
	return container.Contains(*element)
}

func makeSlice(size int) []string {
	slice := make([]string, size)
	for i := 0; i < size-1; i++ {
		slice[i] = "Neo.ClientError.Security.Unauthorized"
	}
	slice[size-1] = "Neo.ClientError.Security.TokenExpired"
	return slice
}

func BenchmarkSliceVsSet(b *testing.B) {
	b.ResetTimer()
	sizes := []int{1, 10, 100, 1000, 10000}
	for _, size := range sizes {
		slice := makeSlice(size)
		set := collections.NewSet[string](slice)
		element := "Neo.ClientError.Security.TokenExpired"

		b.Run(fmt.Sprintf("slice_size_%d", size), func(b *testing.B) {
			for i := 0; i < b.N; i++ {
				fooSlice(&slice, &element)
			}
		})

		b.Run(fmt.Sprintf("set_size_%d", size), func(b *testing.B) {
			for i := 0; i < b.N; i++ {
				fooSet(&set, &element)
			}
		})
	}
}
```
```
goos: linux
goarch: amd64
pkg: github.com/neo4j/neo4j-go-driver/v5/neo4j
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
BenchmarkSliceVsSet
BenchmarkSliceVsSet/slice_size_1
BenchmarkSliceVsSet/slice_size_1-16         	422922162	         2.754 ns/op
BenchmarkSliceVsSet/set_size_1
BenchmarkSliceVsSet/set_size_1-16           	16231747	        73.64 ns/op
BenchmarkSliceVsSet/slice_size_10
BenchmarkSliceVsSet/slice_size_10-16        	23253615	        52.92 ns/op
BenchmarkSliceVsSet/set_size_10
BenchmarkSliceVsSet/set_size_10-16          	 9170173	       138.6 ns/op
BenchmarkSliceVsSet/slice_size_100
BenchmarkSliceVsSet/slice_size_100-16       	 2230754	       536.7 ns/op
BenchmarkSliceVsSet/set_size_100
BenchmarkSliceVsSet/set_size_100-16         	 4424978	       271.3 ns/op
BenchmarkSliceVsSet/slice_size_1000
BenchmarkSliceVsSet/slice_size_1000-16      	  222684	      5338 ns/op
BenchmarkSliceVsSet/set_size_1000
BenchmarkSliceVsSet/set_size_1000-16        	  465009	      2466 ns/op
BenchmarkSliceVsSet/slice_size_10000
BenchmarkSliceVsSet/slice_size_10000-16     	   22353	     53524 ns/op
BenchmarkSliceVsSet/set_size_10000
BenchmarkSliceVsSet/set_size_10000-16       	   63013	     18765 ns/op
```
Bottom line: for short collections, the overhead that a hash set brings seems to outweigh the theoretical runtime complexity benefit compared to a simple slice.

robsdedude: Then again, this is not on the hot path of the driver... So realistically I just wasted my time :joy:

robsdedude: Scratch all of that. I just saw the `Contains` implementation: it's implemented very inefficiently. With the new implementation (see comment), things look very different!

```
goos: linux
goarch: amd64
pkg: github.com/neo4j/neo4j-go-driver/v5/neo4j
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
BenchmarkSliceVsSet
BenchmarkSliceVsSet/slice_size_1
BenchmarkSliceVsSet/slice_size_1-16         	433251711	         2.894 ns/op
BenchmarkSliceVsSet/set_size_1
BenchmarkSliceVsSet/set_size_1-16           	233334042	         5.314 ns/op
BenchmarkSliceVsSet/slice_size_2
BenchmarkSliceVsSet/slice_size_2-16         	137415747	         8.724 ns/op
BenchmarkSliceVsSet/set_size_2
BenchmarkSliceVsSet/set_size_2-16           	170600866	         6.966 ns/op
BenchmarkSliceVsSet/slice_size_5
BenchmarkSliceVsSet/slice_size_5-16         	41603752	        24.63 ns/op
BenchmarkSliceVsSet/set_size_5
BenchmarkSliceVsSet/set_size_5-16           	175438876	         6.858 ns/op
BenchmarkSliceVsSet/slice_size_10
BenchmarkSliceVsSet/slice_size_10-16        	23322840	        51.50 ns/op
BenchmarkSliceVsSet/set_size_10
BenchmarkSliceVsSet/set_size_10-16          	96384018	        11.99 ns/op
BenchmarkSliceVsSet/slice_size_100
BenchmarkSliceVsSet/slice_size_100-16       	 2243829	       534.4 ns/op
BenchmarkSliceVsSet/set_size_100
BenchmarkSliceVsSet/set_size_100-16         	100000000	        11.42 ns/op
BenchmarkSliceVsSet/slice_size_1000
BenchmarkSliceVsSet/slice_size_1000-16      	  222481	      5350 ns/op
BenchmarkSliceVsSet/set_size_1000
BenchmarkSliceVsSet/set_size_1000-16        	100000000	        10.88 ns/op
BenchmarkSliceVsSet/slice_size_10000
BenchmarkSliceVsSet/slice_size_10000-16     	   22423	     53132 ns/op
BenchmarkSliceVsSet/set_size_10000
BenchmarkSliceVsSet/set_size_10000-16       	100000000	        10.89 ns/op
```
That's much more in line with what I expected of `Set.Contains` being somewhere in the ballpark of `O(1)` with some (non-constant) overhead for bigger sets because of cache pages, hash collisions and stuff.

New bottom-line: keep the set, but fix it :)

neo4j/internal/collections/set.go
```diff
@@ -73,3 +73,12 @@ func (set Set[T]) Copy() Set[T] {
...
 }
+
+func (set Set[T]) Contains(value T) bool {
+	for _, a := range set.Values() {
+		if a == value {
+			return true
+		}
+	}
+	return false
+}
```
robsdedude:
```suggestion
func (set Set[T]) Contains(value T) bool {
	_, ok := set[value]
	return ok
}
```

please make use of the "Hash" in `HashMap` for more fast :runner::dash:

neo4j/internal/collections/set_test.go
```diff
@@ -147,6 +147,17 @@ func TestSet(outer *testing.T) {
...
+	outer.Run("contains", func(t *testing.T) {
+		strings := collections.NewSet([]string{
+			"golang",
+			"neo4j",
+		})
+		expected := "golang"
+		if found := strings.Contains(expected); !found {
+			t.Errorf("Set does not contain %v", expected)
+		}
+	})
```
robsdedude:
```suggestion
	outer.Run("contains", func(t *testing.T) {
		strings := collections.NewSet([]string{
			"golang",
			"neo4j",
		})
		expected := "golang"
		if found := strings.Contains(expected); !found {
			t.Errorf("Set should have contained %v", expected)
		}
	})

	outer.Run("does not contain", func(t *testing.T) {
		strings := collections.NewSet([]string{
			"golang",
			"neo4j",
		})
		expected := "foobar"
		if found := strings.Contains(expected); found {
			t.Errorf("Set should not have contain %v", expected)
		}
	})
```

also test negative case

neo4j/auth/auth_examples_test.go
```diff
@@ -28,23 +28,39 @@ import (
 	"time"
 )
 
-func ExampleExpirationBasedTokenManager() {
-	myProvider := func(ctx context.Context) (neo4j.AuthToken, *time.Time, error) {
+func ExampleBasic() {
```
robsdedude: maybe add some explanatory comments?

neo4j/internal/pool/pool.go
```diff
@@ -453,35 +452,36 @@ func (p *Pool) Return(ctx context.Context, c idb.Connection) {
...
-				database = dbSelector.Database()
-			}
-			p.deactivateWriter(connection.ServerName(), database)
+	}
+	if error.IsRetriableCluster() {
+		var database string
+		if dbSelector, ok := connection.(idb.DatabaseSelector); ok {
+			database = dbSelector.Database()
 		}
+		p.deactivateWriter(connection.ServerName(), database)
```
robsdedude: This should probably happen before calling the auth manager as it's infallible. De-facto are the code-paths mutually exclusive so it doesn't really matter. But that might change in the future if `error.IsRetriableCluster()` or `error.HasSecurityCode()` change their semantic.

testkit-backend/backend.go
```diff
@@ -98,9 +98,9 @@ func (g GenericTokenManager) GetAuthToken(_ context.Context) (neo4j.AuthToken, e
 	return g.GetAuthTokenFunc(), nil
 }
 
-func (g GenericTokenManager) OnTokenExpired(_ context.Context, token neo4j.AuthToken) error {
-	g.OnTokenExpiredFunc(token)
-	return nil
+func (g GenericTokenManager) HandleSecurityException(_ context.Context, token neo4j.AuthToken, securityException *db.Neo4jError) (bool, error) {
+	g.HandleSecurityExceptionFunc(token, securityException)
+	return false, nil
```
robsdedude: Feels like this should not just return `false` but actually ask TestKit what to return.

robsdedude (changes requested):

testkit-backend/backend.go
robsdedude, line 1: Also note that the TestKit tests for this feature introduced a new feature flag `API_RETRYABLE_EXCEPTION = "Feature:API:RetryableExceptions"`. For better test coverage, I highly recommend implementing it in the backend.

neo4j/auth/auth.go
StephenCathcart: Agreed, was sticking to ADR but I prefer '...Manager' varients too 👍

neo4j/auth/auth.go
StephenCathcart: I think we should name this to `neo4jAuthTokenManager` too 👍

neo4j/auth/auth.go
StephenCathcart: Yep that's an oversite I just updated the struct name 👍

neo4j/auth/auth_examples_test.go
StephenCathcart: Yep agreed

neo4j/auth/auth.go
StephenCathcart: Thanks for the thorough benchmarking, Contains has been updated to suggested implementation 👍

neo4j/auth/auth.go
StephenCathcart: I've had a bit of a redo of the two factory methods 👍

testkit-backend/backend.go
StephenCathcart: Added the flag in and will update backend in next commit

neo4j/internal/pool/pool.go
StephenCathcart: Makes sense, I've moved the DatabaseUnavailable and IsRetriableCluster checks before the HasSecurityCode check

neo4j/internal/collections/set_test.go
```diff
@@ -147,6 +147,28 @@ func TestSet(outer *testing.T) {
 			t.Error(err)
 		}
 	})
+
+	outer.Run("contains", func(t *testing.T) {
```
fbiville, line 151: nit: I think this one is covered by the first test above, see:
```
outer.Run("adds", func(t *testing.T) {
// [...]
```

testkit-backend/backend.go
StephenCathcart: updated 👍

neo4j/internal/collections/set_test.go
StephenCathcart: Good spot - I like how this one explicitly tests that the set contains a value using the public Contains function rather than relying on the "adds" to capture it but happy to remove either way.

PR #531, opened 2023-09-22
Improve transaction timeout docs

robsdedude:
The timeout behaves differently on different server versions in respects to the user being or not being able to overwrite the server timeout with a bigger value.

Also make sure the docs mention special values like `0`.

### DIFF
--- a/neo4j/transaction_config.go
+++ b/neo4j/transaction_config.go
@@ -35,7 +35,7 @@ type TransactionConfig struct {
 //
 // Transactions that execute longer than the configured timeout will be terminated by the database.
 // This functionality allows user code to limit query/transaction execution time.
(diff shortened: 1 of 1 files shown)


neo4j/transaction_config.go
```diff
@@ -33,16 +33,30 @@ type TransactionConfig struct {
 
 // WithTxTimeout returns a transaction configuration function that applies a timeout to a transaction.
 //
+// Transactions that execute longer than the configured timeout will be terminated by the database.
+// This functionality allows to limit query/transaction execution time.
```
thelonelyvulpes:
```suggestion
// This functionality allows the driver to limit query/transaction execution time.
```

thelonelyvulpes (approved):

StephenCathcart (approved):

PR #533, opened 2023-09-27
Add Bolt 5.4 support and Telemetry API

StephenCathcart:
Neo4j would like to be able to track more information about driver API usage so that smarter decisions can be made on improving drivers and their stack.

The collection of metrics must be respectful of users. Thus, the collected metrics are:
1. impossible to be tied back to the customer or user
1. transparent to the users that inspect
1. restrained to required metrics

The metric collection is disabled by default in Neo4j. It can be enabled in the server by setting `server.bolt.telemetry.enabled` to `true`.

However, the metric collection is enabled by default in the drivers. It can be disabled in the driver by configuring the driver with `TelemetryDisabled` to `true`. Default: `false`.

**Metrics are only collected when enabled both in server and driver instances.**

Disabling metrics on driver:
```go
driver, _ := neo4j.NewDriverWithContext("neo4j://localhost:7687", neo4j.BasicAuth("neo4j", "password", ""), func(config *config.Config) {
    config.TelemetryDisabled = true
})
```

### DIFF
--- a/neo4j/config.go
+++ b/neo4j/config.go
@@ -50,6 +50,7 @@ func defaultConfig() *Config {
 		FetchSize:                       FetchDefault,
 		NotificationsMinSeverity:        notifications.DefaultLevel,
 		NotificationsDisabledCategories: notifications.NotificationDisabledCategories{},
+		TelemetryDisabled:               false,
 	}
 }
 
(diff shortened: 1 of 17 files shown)


StephenCathcart:
[Metrics Implementation [Go]](https://trello.com/c/PnpfTac0/28-metrics-implementation-go)

gjmwoods (approved):
💪

PR #536, opened 2023-10-10
Fix transaction termination handling

StephenCathcart:
* Added a `transactionState` struct to hold error and callbacks that are shared with `Results`.
* Fixed bug in `Commit` by checking for `tx.conn.HasFailed()` before returning `transactionAlreadyCompletedError`.
* Removed dependency on `runFailed` variable by deleting it and replacing it with checks from `transactionState` error.
* Results throw a `result failed` error when trying to interact with a result that came from a transaction that errored.


StephenCathcart:
[Fix transaction handling [Go]](https://trello.com/c/3Jep6Gs9/62-fix-transaction-handling-go)

neo4j/session_with_context.go
```diff
@@ -352,8 +352,11 @@ func (s *sessionWithContext) BeginTransaction(ctx context.Context, configurers .
 			s.pool.Return(ctx, tx.conn)
 			tx.err = errorutil.CombineAllErrors(tx.err, bookmarkErr)
 			tx.conn = nil
+			// TODO not sure if clearing the results here is necessary.
+			tx.results = nil
 			s.explicitTx = nil
 		},
+		results: make([]ResultWithContext, 0),
```
fbiville: disclaimer: I'm not sure this will make the code better

it's likely going to be a bit more lightweight to store a slice of `func(error)`
assuming the change looks like:
```go
resultCallbacks : make([]func(error))
```

the tx code below would become:
```go
	for _, resultCallback := range tx.resultCallbacks {
			// TODO possibly replace err with the a new 'ResultInvalid' error
			resultCallback(err)
		}
```
and
```go
tx.resultCallbacks = append(tx.resultCallbacks, result.onTxError)
```

robsdedude: Alternative: have a struct like `TransactionState` that is shared between the tx and its results. State could be one of `Open`, `Closed`, `Failed` with some `err`.

neo4j/transaction_with_context.go
```diff
@@ -70,26 +88,23 @@ func (tx *explicitTransaction) Run(ctx context.Context, cypher string, params ma
...
+	return result, nil
 }
 
 func (tx *explicitTransaction) Commit(ctx context.Context) error {
-	if tx.runFailed {
-		tx.runFailed = false
-		return tx.err
-	}
-	if tx.conn == nil {
+	if tx.conn == nil || tx.conn.HasFailed() {
```
StephenCathcart: `tx.conn.HasFailed()` was missing here causing `test_should_prevent_commit_after_tx_termination` to fail.

neo4j/transaction_with_context.go
```diff
@@ -70,26 +88,23 @@ func (tx *explicitTransaction) Run(ctx context.Context, cypher string, params ma
 	}
 	stream, err := tx.conn.RunTx(ctx, tx.txHandle, db.Command{Cypher: cypher, Params: params, FetchSize: tx.fetchSize})
 	if err != nil {
-		tx.err = err
-		tx.runFailed = true
+		tx.txState.onError(err)
```
StephenCathcart, line 84: fixes `test_should_prevent_discard_after_tx_termination_on_run due` and `test_should_prevent_pull_after_tx_termination_on_run` by propagating errors

neo4j/driver_with_context_test.go
```diff
@@ -710,6 +710,10 @@ func (f *fakeResult) legacy() Result {
 	panic("implement me")
 }
 
+func (f *fakeResult) errorHandler(err error) {
```
robsdedude:
```suggestion
func (f *fakeResult) errorHandler(error) {
```
:broom:

neo4j/result_with_context.go
```diff
@@ -72,15 +75,23 @@ type resultWithContext struct {
...
 }
 
-func newResultWithContext(connection idb.Connection, stream idb.StreamHandle, cypher string, params map[string]any, afterConsumptionHook func()) ResultWithContext {
+func newResultWithContext(
+	connection idb.Connection,
+	stream idb.StreamHandle,
+	cypher string,
+	params map[string]any,
+	txState *transactionState,
+	afterConsumptionHook func()) ResultWithContext {
```
robsdedude:
```suggestion
	afterConsumptionHook func(),
) ResultWithContext {
```
Just a personal preference. Otherwise, I find it quite hard to see where the parameter list ends and the function body starts.

neo4j/transaction_with_context.go
robsdedude: should `|| !tx.conn.IsAlive()` also be checked as is done in `Rollback()`?

robsdedude (commented):
This looks really good already. But I don't see, how a failing record (encountering an error while streaming), would propagate to the other results in the same transaction.

neo4j/transaction_with_context.go
StephenCathcart: I think so, I've re-ran Testkit with this change and all seems good 👍.

neo4j/transaction_with_context.go
```diff
@@ -54,14 +54,32 @@ type ExplicitTransaction interface {
...
 
+type transactionState struct {
+	err                 error
+	resultErrorHandlers []func(error)
+}
+
+func newTransactionState() *transactionState {
+	return &transactionState{
+		err:                 nil,
+		resultErrorHandlers: make([]func(error), 0),
```
robsdedude: More for my education: is this for all we're doing with it equivalent?
```suggestion
		resultErrorHandlers: nil,
```

neo4j/transaction_with_context.go
```diff
@@ -70,26 +88,26 @@ func (tx *explicitTransaction) Run(ctx context.Context, cypher string, params ma
 	}
 	stream, err := tx.conn.RunTx(ctx, tx.txHandle, db.Command{Cypher: cypher, Params: params, FetchSize: tx.fetchSize})
 	if err != nil {
-		tx.err = err
-		tx.runFailed = true
-		tx.onClosed(tx)
-		return nil, errorutil.WrapError(tx.err)
+		tx.txState.onError(err)
+		tx.onClosed()
```
robsdedude:
```suggestion
```
Do we still need to call this, considering that session will register onClose to be called on `txState.onError`? It my assumption true for managed transactions as well?

neo4j/transaction_with_context.go
StephenCathcart: The `make` will allocate and return a non-nil slice of length and capacity 0 rather than having a nil value. However, in our case using nil will be fine as both `append` and iterate work on nil slices so this is now redundant. We can get rid of the newTransaction constructor as this was created before the large refactor where we needed the slice to not be nil. I'll have a refactor 👍

neo4j/transaction_with_context.go
StephenCathcart: No, we can remove it, it gets closed when the `onError` is called 👍

robsdedude (approved):
:artificial_satellite:

PR #551, opened 2023-11-30
Add `ConnectionLivenessCheckTimeout` configuration

StephenCathcart:
This configuration defines the duration that the connection may be idle before needing to perform a liveness check on acquiring from the pool.

```golang
driver, err := neo4j.NewDriverWithContext("neo4j://localhost:7687", neo4j.BasicAuth("neo4j", "password", ""), func(config *neo4j.Config) {
    config.ConnectionLivenessCheckTimeout = 2 * time.Second
})
```

Check the API docs for more information.

### DIFF
--- a/neo4j/config.go
+++ b/neo4j/config.go
@@ -19,6 +19,7 @@ package neo4j
 
 import (
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/config"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/pool"
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/notifications"
 	"math"
 	"net/url"
@@ -41,6 +42,7 @@ func defaultConfig() *Config {
 		MaxConnectionPoolSize:           100,
 		MaxConnectionLifetime:           1 * time.Hour,
 		ConnectionAcquisitionTimeout:    1 * time.Minute,
+		ConnectionLivenessCheckTimeout:  pool.DefaultConnectionLivenessCheckTimeout,
 		SocketConnectTimeout:            5 * time.Second,
 		SocketKeepalive:                 true,
 		RootCAs:                         nil,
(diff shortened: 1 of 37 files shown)


StephenCathcart:
[[go] Implement ADR 021: liveness check timeout](https://trello.com/c/9jTJ31i4/124-go-implement-adr-021-liveness-check-timeout)

neo4j/config.go
```diff
@@ -77,6 +79,11 @@ func validateAndNormaliseConfig(config *Config) error {
 		config.ConnectionAcquisitionTimeout = -1
 	}
 
+	// Connection Liveness Check Timeout
+	if config.ConnectionLivenessCheckTimeout < 0 {
+		return &UsageError{Message: "Connection liveness check timeout cannot be smaller than 0"}
+	}
```
robsdedude, line 85: Consider adding a unit test for this case (there already is a `TestValidateAndNormaliseConfig` that you can extend).

neo4j/internal/router/router_test.go
```diff
@@ -56,7 +57,7 @@ func TestMultithreading(t *testing.T) {
 		n = n.Add(time.Duration(table.TimeToLive) * time.Second * 2)
 		return n
 	}
-	router := New("router", func() []string { return []string{} }, nil, pool, logger, "routerid", &timer)
+	router := New("router", func() []string { return []string{} }, nil, pool, math.MaxInt64, logger, "routerid", &timer)
```
robsdedude: Any reasons for/against using `DefaultConnectionLivenessCheckTimeout` here instead?

robsdedude (approved):
You might want to add a TestKit test skip in accordance with the decision documented in [the ADR update (internal link)](https://github.com/neo-technology/drivers-adr/pull/83). Something along the lines of [this](https://github.com/neo4j/neo4j-python-driver/commit/20126112dde1b82da0c441c0232ff756f0830b5b).

neo4j/config.go
StephenCathcart: Yep agree 👍

PR #552, opened 2023-12-07
Non-racing lock for driver pool

robsdedude:
The mutex guarding the driver's pool attribute does not need to be held while doing IO. Therefore, we can use a standard (non-cancelable) mutex to guard it. This comes at the benefit of not needing a context for session creation. Session creation doesn't perform any I/O, so it shouldn't take long anyway. There's little to no gain in it accepting a context.


StephenCathcart:
lock updates make sense to me 👍

StephenCathcart (approved):

PR #554, opened 2023-12-08
Minor code clean-ups

robsdedude:


neo4j/internal/pool/pool.go
```diff
@@ -284,10 +291,13 @@ func (p *Pool) Borrow(ctx context.Context, getServerNames func() []string, wait
 	}
 }
 
-func (p *Pool) tryBorrow(ctx context.Context, serverName string, boltLogger log.BoltLogger, idlenessThreshold time.Duration, auth *idb.ReAuthToken) (idb.Connection, error) {
-	// For now, lock complete servers map to avoid over connecting but with the downside
```
StephenCathcart, line 283: Are these comments not correct anymore?

StephenCathcart (approved):
🚀

neo4j/internal/pool/pool.go
robsdedude: They are outdated.

With https://github.com/neo4j/neo4j-go-driver/pull/467 the mechanism for preventing over connecting was changed to counting reservations. It also changed it so that no connection is established (and more generally no I/O is happening) while the lock is being held.

PR #564, opened 2024-01-29
Add  `ExecuteQueryWithAuthToken` configuration option

StephenCathcart:
```golang
result, err := neo4j.ExecuteQuery(ctx, driver, query, parameters,
    neo4j.ExecuteQueryWithAuthToken(neo4j.BasicAuth("neo4j", "password", "realm")),
)
```

### DIFF
--- a/neo4j/driver_with_context.go
+++ b/neo4j/driver_with_context.go
@@ -672,6 +672,13 @@ func ExecuteQueryWithTransactionConfig(configurers ...func(*TransactionConfig)) 
 	}
 }
 
+// ExecuteQueryWithAuthToken configures DriverWithContext.ExecuteQuery to overwrite the AuthToken for the session.
+func ExecuteQueryWithAuthToken(auth AuthToken) ExecuteQueryConfigurationOption {
+	return func(configuration *ExecuteQueryConfiguration) {
+		configuration.Auth = &auth
+	}
+}
+
 // ExecuteQueryConfiguration holds all the possible configuration settings for DriverWithContext.ExecuteQuery
 type ExecuteQueryConfiguration struct {
 	Routing                RoutingControl
@@ -680,6 +687,7 @@ type ExecuteQueryConfiguration struct {
 	BookmarkManager        BookmarkManager
 	BoltLogger             log.BoltLogger
 	TransactionConfigurers []func(*TransactionConfig)
+	Auth                   *AuthToken
 }
 
 // RoutingControl specifies how the query executed by DriverWithContext.ExecuteQuery is to be routed
@@ -698,6 +706,7 @@ func (c *ExecuteQueryConfiguration) toSessionConfig() SessionConfig {
 		DatabaseName:     c.Database,
 		BookmarkManager:  c.BookmarkManager,
 		BoltLogger:       c.BoltLogger,
+		Auth:             c.Auth,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


StephenCathcart:
[[Go] Expose session auth on driver.executeQuery](https://trello.com/c/sGtMX1b7/190-go-expose-session-auth-on-driverexecutequery)

testkit-backend/backend.go
```diff
@@ -600,6 +600,12 @@ func (b *backend) handleRequest(req map[string]any) {
 				if executeQueryConfig["txMeta"] != nil {
 					config.TransactionConfigurers = append(config.TransactionConfigurers, neo4j.WithTxMetadata(b.toTxMetadata(executeQueryConfig)))
 				}
+				// Append Auth configuration if it exists
+				if executeQueryConfig["auth"] != nil {
+					auth := executeQueryConfig["auth"].(map[string]any)
+					token := neo4j.BasicAuth(auth["principal"].(string), auth["credentials"].(string), "")
+					config.Auth = &token
+				}
```
robsdedude: This can only deserialize basic auth tokens. I think it'd be good if TestKit had the liberty to send any type of auth token (like those encoded in the `NewDriver` message). There's already a `getAuth` function in the Go driver's TestKit backend to deserialize different types of auth tokens.

neo4j/driver_with_context.go
```diff
@@ -672,6 +672,13 @@ func ExecuteQueryWithTransactionConfig(configurers ...func(*TransactionConfig))
 	}
 }
 
+// ExecuteQueryWithAuthToken configures DriverWithContext.ExecuteQuery to overwrite the AuthToken for the session.
+func ExecuteQueryWithAuthToken(auth AuthToken) ExecuteQueryConfigurationOption {
```
robsdedude, line 676: This is up for discussion:
Why name it `ExecuteQueryWithAuthToken` over `ExecuteQueryWithAuth`, given that the field on `ExecuteQueryConfiguration` (as well as the field on `SessionConfig`) is called `Auth`, not `AuthToken`?

StephenCathcart: My thinking was that it's because you use an `AuthToken` to authenticate (like you do in `VerifyAuthentication(ctx context.Context, auth *AuthToken...)` and not an `auth.TokenManager` (like you do in `NewDriverWithContext(target string, auth auth.TokenManager...`) to avoid any confusion for the user.

We seem to call both types `auth` internally:
driverWithContext struct:
`auth auth.TokenManager`

SessionConfig:
`Auth *AuthToken`

To keep the API consistent thought I'm happy to rename it just to auth.

robsdedude: Good argument you've got there :thinking:
I assume the fact that `TokenManager` was an afterthought already blurs the lines here as session config doesn't accept a manager, but the field is still called `Auth`.

Yeah, I don't know :upside_down_face: I'll leave it up to you to decide how to name this method.

testkit-backend/backend.go
StephenCathcart: Thanks for pointing this out I didn't realise it existed 👍 I've updated the backend to use `getAuth` instead.

testkit-backend/backend.go
```diff
@@ -561,6 +561,9 @@ func (b *backend) handleRequest(req map[string]any) {
 		if rawConfig := data["config"]; rawConfig != nil {
 			executeQueryConfig := rawConfig.(map[string]any)
 			configurers = append(configurers, func(config *neo4j.ExecuteQueryConfiguration) {
+
```
robsdedude:
```suggestion
```

robsdedude (approved):
:name_badge:

PR #581, opened 2024-05-12
Buffered reading.

thelonelyvulpes:
This change introduces a buffered reader between the racing reader and the connection to reduce the number of syscalls required to read messages.

The size of the read buffer is configurable on the driver.
- Each connection allocates an individual buffer, so memory usage could increase, users should configure this where appropriate, if more syscalls for network reads is seen then increasing the buffer size could help.

Assuming a 16KB Buffer for 100 connections on 3 servers that would equal 16KB * 100 * 3 = ~5MB.
N.b. that 300 connections would indicate a very busy client and 5MB should be considered a relatively small increased footprint.

The default read buffer size is: 32K?

### DIFF
--- a/neo4j/config.go
+++ b/neo4j/config.go
@@ -19,8 +19,8 @@ package neo4j
 
 import (
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j/config"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/bolt"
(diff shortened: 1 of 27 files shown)


StephenCathcart:
~~Tests failing due to `bufio.Reader` not being thread-safe.~~

Tests fixed

StephenCathcart (approved):

robsdedude (approved):
:basket:

PR #586, opened 2024-06-18
Clarify `FetchAll` behavior inline docs

stefano-ottolenghi:
"single large result" is ambiguous, as it refer both to a session used to fetch _one result_ with many records, or  to a session used to fetch many results, each with a single record.

### DIFF
--- a/neo4j/config/driver.go
+++ b/neo4j/config/driver.go
@@ -168,10 +168,10 @@ type Config struct {
 	// FetchSize defines how many records to pull from server in each batch.
 	// From Bolt protocol v4 (Neo4j 4+) records can be fetched in batches as
 	// compared to fetching all in previous versions.
+	// If the underlying protocol does not support FetchSize, it is ignored.
 	//
 	// If FetchSize is set to `FetchDefault`, the driver decides the appropriate size.
-	// If set to a positive value that size is used (if the underlying protocol supports it);
-	// otherwise it is ignored.
+	// If set to a positive value that size is used; negative values behaves as `FetchAll` (see below).
 	//
 	// To turn off fetching in batches and always fetch everything, set FetchSize to `FetchAll`.
 	// A limited FetchSize ensures the client is not overflown with records,
--- a/neo4j/config/driver.go
+++ b/neo4j/config/driver.go
@@ -169,14 +169,13 @@ type Config struct {
 	// From Bolt protocol v4 (Neo4j 4+) records can be fetched in batches as
 	// compared to fetching all in previous versions.
 	//
-	// If FetchSize is set to FetchDefault, the driver decides the appropriate
+	// If FetchSize is set to `FetchDefault`, the driver decides the appropriate
 	// size. If set to a positive value that size is used if the underlying
 	// protocol supports it otherwise it is ignored.
 	//
 	// To turn off fetching in batches and always fetch everything, set
-	// FetchSize to FetchAll.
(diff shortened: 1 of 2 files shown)


neo4j/config/driver.go
```diff
@@ -169,14 +169,13 @@ type Config struct {
...
+	// If FetchSize is set to `FetchDefault`, the driver decides the appropriate
 	// size. If set to a positive value that size is used if the underlying
 	// protocol supports it otherwise it is ignored.
 	//
 	// To turn off fetching in batches and always fetch everything, set
-	// FetchSize to FetchAll.
-	// If a single large result is to be retrieved, this is the most performant
-	// setting.
+	// FetchSize to `FetchAll`.
+	// `FetchAll` is the most performant setting for results consisting of a single large record.
```
robsdedude: For a single record, the fetch size doesn't matter. More generally, any fetch size <= number of records will result in the lowest possible numbers of round trips and smallest possible bandwidth required given the current design of the bolt protocol.

Maybe something along the lines of `FetchAll` is the most performant setting at the cost of increased memory requirements.

On a more general note, anything on the spectrum from fetch size 1 to `FetchAll` is moving the line of that trade-off. Small fetch size => small memory footprint (server and client side)  but longer runtime and vice-versa.

stefano-ottolenghi: The remark smelled bad to me too, but I assumed it would be right given that it was there :facepalm: I thought _maybe_ `PULL n` (with `n != -1`) would also take a few extra ending messages to confirm that records are over, rather than the client knowing it ahead of time.

> Maybe something along the lines of FetchAll is the most performant setting at the cost of increased memory requirements.

That's not true either. The most efficient setting is a fetch size that quickly provides a fair amount of records to the client, which can process them while the server fetches _more_ records. That size is different for each query, as it depends how fast the server can fetch records, and also how large they are to transmit. 

At this point I think this `FetchAll` remark is best removed.
The whole paragraph could use some cleanup and should focus on the tradeoff you mentioned, which is what users really should know about this setting.
Any objections to this?

PS. An side, but very important.
We've detailed how it _should_ work, but it's not what I found in my tests to craft the new [lazy-eager-loading example](https://neo4j.com/docs/go-manual/current/performance/#lazy-eager-loading) (which is why I've avoided mentioning the fetch size there). In my tests, the driver always _streamed_ records, regardless of fetch size (I've tried different magnitudes, I think ~ 10, 100, 1000). No matter what fetch size I set, the first record would always be available very quickly, and subsequent records would come with the same delay. From a user perspective, it behaved as if fetch size was always `1`.

neo4j/config/driver.go
```diff
@@ -169,13 +169,13 @@ type Config struct {
 	// From Bolt protocol v4 (Neo4j 4+) records can be fetched in batches as
 	// compared to fetching all in previous versions.
 	//
-	// If FetchSize is set to `FetchDefault`, the driver decides the appropriate
-	// size. If set to a positive value that size is used if the underlying
-	// protocol supports it otherwise it is ignored.
+	// If FetchSize is set to `FetchDefault`, the driver decides the appropriate size.
+	// If set to a positive value that size is used (if the underlying protocol supports it);
+	// otherwise it is ignored.
```
robsdedude:
```suggestion
	// If set to a positive value that size is used (if the underlying protocol supports it;
	// otherwise it is ignored).
```

robsdedude (approved):

PR #587, opened 2024-06-19
Switch to Go 1.19 baseline

sp71:
- Upgrade Go to 1.19 from 1.18
- Lack access to TeamCity, feel free to take over the PR and push commits


StephenCathcart:
Hi, @sp71, thanks for taking the time to open a PR for the Go driver!

Unfortunately, we don't bump Go versions outside of a major release to avoid breaking changes in client code. The next major bump for us will be the 6.0 driver. We'll bump the Go version to the latest available at that time.

PR #588, opened 2024-06-26
Add `GqlStatusObject` support as notifications to `ResultSummary`

StephenCathcart:
⚠️  This API is released as preview.

Introduces GQL-compliant status objects to the `ResultSummary`.

`GqlStatusObject` is a GQL-compliant `Notification` object and status of query execution, this new object includes `GqlStatus` and `StatusDescription`.

`ResultSummary.GqlStatusObjects()` always contains at least 1 status representing the `Success`, `No Data` or `Omitted Result`.

The GqlStatusObjects will be presented in the following order:

* A “no data” (`02xxx`) has precedence over a warning;
* A warning (`01xxx`) has precedence over a success.
* A success (`00xxx`) has precedence over anything informational (`03xxx`)

**Migration from Notification**

Most properties in `Notification` exist in the new `GqlStatusObject`, except `Code` and `Title`.
The `GqlStatusObject.GqlStatus` is equivalent to `Code` in `Notification`, but with values compliant with GQL.
The properties `GqlStatusObject.Classification` and `GqlStatusObject.RawClassification` are equivalent to `Notification.Category` and `Notification.RawCategory`. The name change is needed because Category has a different meaning in GQL.

**Usage**

Disabling GqlStatusObjects can be done in the same way as disabling Notifications. However, non-notification status can not be filtered (such as Success or No Data):

```go
neo4j.NewDriverWithContext("neo4j://localhost", neo4j.BasicAuth("neo4j", "password", ""), func(config *config.Config) {
    config.NotificationsDisabledClassifications = notifications.DisableClassifications(notifications.Deprecation, notifications.Performance)
})
```

Iterating through GqlStatusObjects is also done in the same way as Notifications:

```go
gqlStatusObjects := result.Summary.GqlStatusObjects()
for _, gqlStatusObject := range gqlStatusObjects {
    fmt.Printf(gqlStatusObject.GqlStatus())
    fmt.Printf(gqlStatusObject.StatusDescription())
    fmt.Printf(gqlStatusObject.DiagnosticRecord())
    // and so on
}
``` 

**Deprecations**

- `neo4j.NotificationSeverity` please use `notifications.NotificationSeverity` directly.
- `neo4j.Warning` please use `notifications.Warning` directly.
- `neo4j.Information` please use `notifications.Information` directly.
- `neo4j.UnknownSeverity` please use `notifications.UnknownSeverity` directly.

- `neo4j.NotificationCategory` please use `notifications.NotificationCategory` directly.
- `neo4j.Hint` please use `notifications.Hint` directly.
- `neo4j.Unrecognized` please use `notifications.Unrecognized` directly.
- `neo4j.Unsupported` please use `notifications.Unsupported` directly.
- `neo4j.Performance` please use `notifications.Performance` directly.
- `neo4j.Deprecation` please use `notifications.Deprecation` directly.
- `neo4j.Generic` please use `notifications.Generic` directly.
- `neo4j.Security` please use `notifications.Security` directly.
- `neo4j.Topology` please use `notifications.Topology` directly.
- `neo4j.UnknownCategory` please use `notifications.Unknown` directly.

### DIFF
--- a/neo4j/config.go
+++ b/neo4j/config.go
@@ -37,20 +37,21 @@ type ServerAddress = config.ServerAddress
 
 func defaultConfig() *Config {
 	return &Config{
-		AddressResolver:                 nil,
-		MaxTransactionRetryTime:         30 * time.Second,
-		MaxConnectionPoolSize:           100,
-		MaxConnectionLifetime:           1 * time.Hour,
-		ConnectionAcquisitionTimeout:    1 * time.Minute,
-		ConnectionLivenessCheckTimeout:  pool.DefaultConnectionLivenessCheckTimeout,
-		SocketConnectTimeout:            5 * time.Second,
-		SocketKeepalive:                 true,
-		RootCAs:                         nil,
-		UserAgent:                       UserAgent,
-		FetchSize:                       FetchDefault,
-		NotificationsMinSeverity:        notifications.DefaultLevel,
-		NotificationsDisabledCategories: notifications.NotificationDisabledCategories{},
-		TelemetryDisabled:               false,
+		AddressResolver:                      nil,
+		MaxTransactionRetryTime:              30 * time.Second,
+		MaxConnectionPoolSize:                100,
+		MaxConnectionLifetime:                1 * time.Hour,
+		ConnectionAcquisitionTimeout:         1 * time.Minute,
+		ConnectionLivenessCheckTimeout:       pool.DefaultConnectionLivenessCheckTimeout,
+		SocketConnectTimeout:                 5 * time.Second,
+		SocketKeepalive:                      true,
+		RootCAs:                              nil,
+		UserAgent:                            UserAgent,
+		FetchSize:                            FetchDefault,
+		NotificationsMinSeverity:             notifications.DefaultLevel,
+		NotificationsDisabledCategories:      notifications.NotificationDisabledCategories{},
+		NotificationsDisabledClassifications: notifications.NotificationDisabledClassifications{},
+		TelemetryDisabled:                    false,
 	}
 }
 
@@ -87,6 +88,12 @@ func validateAndNormaliseConfig(config *Config) error {
 	// Socket Connect Timeout
 	if config.SocketConnectTimeout < 0 {
 		config.SocketConnectTimeout = 0
+	}
+
+	// Check notifications have not been configured with both categories and classifications.
+	if len(config.NotificationsDisabledCategories.DisabledCategories()) > 0 &&
+		len(config.NotificationsDisabledClassifications.DisabledClassifications()) > 0 {
+		return &UsageError{Message: "Notifications cannot be disabled for both categories and classifications at the same time."}
 	}
 
 	return nil
--- a/neo4j/config.go
+++ b/neo4j/config.go
@@ -38,23 +38,22 @@ type ServerAddress = config.ServerAddress
 
 func defaultConfig() *Config {
 	return &Config{
-		AddressResolver:                nil,
-		MaxTransactionRetryTime:        30 * time.Second,
-		MaxConnectionPoolSize:          100,
-		MaxConnectionLifetime:          1 * time.Hour,
-		ConnectionAcquisitionTimeout:   1 * time.Minute,
-		ConnectionLivenessCheckTimeout: pool.DefaultConnectionLivenessCheckTimeout,
-		SocketConnectTimeout:           5 * time.Second,
-		SocketKeepalive:                true,
-		RootCAs:                        nil,
-		UserAgent:                      UserAgent,
-		FetchSize:                      FetchDefault,
-		NotificationsMinSeverity:       notifications.DefaultLevel,
-		//lint:ignore SA1019 NotificationsDisabledCategories is supported at least until 6.0
+		AddressResolver:                      nil,
+		MaxTransactionRetryTime:              30 * time.Second,
+		MaxConnectionPoolSize:                100,
+		MaxConnectionLifetime:                1 * time.Hour,
+		ConnectionAcquisitionTimeout:         1 * time.Minute,
+		ConnectionLivenessCheckTimeout:       pool.DefaultConnectionLivenessCheckTimeout,
+		SocketConnectTimeout:                 5 * time.Second,
+		SocketKeepalive:                      true,
+		RootCAs:                              nil,
+		UserAgent:                            UserAgent,
+		FetchSize:                            FetchDefault,
+		NotificationsMinSeverity:             notifications.DefaultLevel,
 		NotificationsDisabledCategories:      notifications.NotificationDisabledCategories{},
 		NotificationsDisabledClassifications: notifications.NotificationDisabledClassifications{},
 		TelemetryDisabled:                    false,
-		ReadBufferSize:                  bolt.DefaultReadBufferSize,
+		ReadBufferSize:                       bolt.DefaultReadBufferSize,
 	}
 }
 
@@ -94,7 +93,6 @@ func validateAndNormaliseConfig(config *Config) error {
 	}
 
 	// Check notifications have not been configured with both categories and classifications.
-	//lint:ignore SA1019 NotificationsDisabledCategories and DisabledCategories are supported at least until 6.0
 	if len(config.NotificationsDisabledCategories.DisabledCategories()) > 0 &&
 		len(config.NotificationsDisabledClassifications.DisabledClassifications()) > 0 {
 		return &UsageError{Message: "Notifications cannot be disabled for both categories and classifications at the same time."}
--- a/neo4j/config.go
+++ b/neo4j/config.go
@@ -37,18 +37,19 @@ type ServerAddress = config.ServerAddress
 
 func defaultConfig() *Config {
 	return &Config{
-		AddressResolver:                      nil,
-		MaxTransactionRetryTime:              30 * time.Second,
-		MaxConnectionPoolSize:                100,
-		MaxConnectionLifetime:                1 * time.Hour,
-		ConnectionAcquisitionTimeout:         1 * time.Minute,
-		ConnectionLivenessCheckTimeout:       pool.DefaultConnectionLivenessCheckTimeout,
-		SocketConnectTimeout:                 5 * time.Second,
-		SocketKeepalive:                      true,
-		RootCAs:                              nil,
-		UserAgent:                            UserAgent,
-		FetchSize:                            FetchDefault,
-		NotificationsMinSeverity:             notifications.DefaultLevel,
+		AddressResolver:                nil,
+		MaxTransactionRetryTime:        30 * time.Second,
+		MaxConnectionPoolSize:          100,
+		MaxConnectionLifetime:          1 * time.Hour,
+		ConnectionAcquisitionTimeout:   1 * time.Minute,
+		ConnectionLivenessCheckTimeout: pool.DefaultConnectionLivenessCheckTimeout,
+		SocketConnectTimeout:           5 * time.Second,
+		SocketKeepalive:                true,
+		RootCAs:                        nil,
+		UserAgent:                      UserAgent,
+		FetchSize:                      FetchDefault,
+		NotificationsMinSeverity:       notifications.DefaultLevel,
+		//lint:ignore SA1019 NotificationsDisabledCategories is supported at least until 6.0
 		NotificationsDisabledCategories:      notifications.NotificationDisabledCategories{},
 		NotificationsDisabledClassifications: notifications.NotificationDisabledClassifications{},
 		TelemetryDisabled:                    false,
@@ -91,6 +92,7 @@ func validateAndNormaliseConfig(config *Config) error {
 	}
 
 	// Check notifications have not been configured with both categories and classifications.
+	//lint:ignore SA1019 NotificationsDisabledCategories and DisabledCategories are supported at least until 6.0
 	if len(config.NotificationsDisabledCategories.DisabledCategories()) > 0 &&
 		len(config.NotificationsDisabledClassifications.DisabledClassifications()) > 0 {
 		return &UsageError{Message: "Notifications cannot be disabled for both categories and classifications at the same time."}
--- a/neo4j/config/driver.go
+++ b/neo4j/config/driver.go
@@ -182,10 +182,18 @@ type Config struct {
 	// By default, the server's settings are used.
 	// Disabling severities allows the server to skip analysis for those, which can speed up query execution.
 	NotificationsMinSeverity notifications.NotificationMinimumSeverityLevel
+	// Deprecated: please use NotificationsDisabledClassifications. This will be removed in 6.0.
+	//
 	// NotificationsDisabledCategories defines the categories of notifications the server should not send.
 	// By default, the server's settings are used.
 	// Disabling categories allows the server to skip analysis for those, which can speed up query execution.
 	NotificationsDisabledCategories notifications.NotificationDisabledCategories
+	// NotificationsDisabledClassifications is identical to NotificationsDisabledCategories.
+	// This alternative is provided for a consistent naming with neo4j.GqlStatusObject Classification.
+	//
+	// NotificationsDisabledClassifications is part of the GQL compliant notifications preview feature
+	// (see README on what it means in terms of support and compatibility guarantees)
+	NotificationsDisabledClassifications notifications.NotificationDisabledClassifications
 	// By default, if the server requests it, the driver will automatically transmit anonymous usage
 	// statistics to the server it is connected to.
 	//
--- a/neo4j/config/driver.go
+++ b/neo4j/config/driver.go
@@ -182,13 +182,9 @@ type Config struct {
 	// By default, the server's settings are used.
 	// Disabling severities allows the server to skip analysis for those, which can speed up query execution.
 	NotificationsMinSeverity notifications.NotificationMinimumSeverityLevel
-	// Deprecated: please use NotificationsDisabledClassifications. This will be removed in 6.0.
-	//
 	// NotificationsDisabledCategories defines the categories of notifications the server should not send.
 	// By default, the server's settings are used.
 	// Disabling categories allows the server to skip analysis for those, which can speed up query execution.
-	//
-	//lint:ignore SA1019 NotificationsDisabledCategories is supported at least until 6.0
 	NotificationsDisabledCategories notifications.NotificationDisabledCategories
 	// NotificationsDisabledClassifications is identical to NotificationsDisabledCategories.
 	// This alternative is provided for a consistent naming with neo4j.GqlStatusObject Classification.
--- a/neo4j/config/driver.go
+++ b/neo4j/config/driver.go
@@ -187,6 +187,8 @@ type Config struct {
 	// NotificationsDisabledCategories defines the categories of notifications the server should not send.
 	// By default, the server's settings are used.
 	// Disabling categories allows the server to skip analysis for those, which can speed up query execution.
+	//
+	//lint:ignore SA1019 NotificationsDisabledCategories is supported at least until 6.0
 	NotificationsDisabledCategories notifications.NotificationDisabledCategories
 	// NotificationsDisabledClassifications is identical to NotificationsDisabledCategories.
 	// This alternative is provided for a consistent naming with neo4j.GqlStatusObject Classification.
--- a/neo4j/config_test.go
+++ b/neo4j/config_test.go
@@ -18,6 +18,7 @@
 package neo4j
 
 import (
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/notifications"
 	"math"
 	"testing"
 	"time"
@@ -119,6 +120,18 @@ func TestValidateAndNormaliseConfig(rt *testing.T) {
 		}
 		if config.SocketConnectTimeout != 0*time.Nanosecond {
 			t.Errorf("SocketConnectTimeout should be set to (0 * time.Nanosecond) when negative")
+		}
+	})
+
+	rt.Run("Configure both NotificationsDisabledCategories and NotificationsDisabledCategories", func(t *testing.T) {
+		config := defaultConfig()
+
+		config.NotificationsDisabledCategories = notifications.DisableCategories(notifications.Deprecation)
+		config.NotificationsDisabledClassifications = notifications.DisableClassifications(notifications.Deprecation)
+
+		err := validateAndNormaliseConfig(config)
+		if err == nil {
+			t.Errorf("NotificationsDisabledCategories and NotificationsDisabledCategories are configured but no error returned")
 		}
 	})
 }
--- a/neo4j/config_test.go
+++ b/neo4j/config_test.go
@@ -126,7 +126,6 @@ func TestValidateAndNormaliseConfig(rt *testing.T) {
 	rt.Run("Configure both NotificationsDisabledCategories and NotificationsDisabledCategories", func(t *testing.T) {
 		config := defaultConfig()
 
-		//lint:ignore SA1019 NotificationsDisabledCategories and DisableCategories are supported at least until 6.0
 		config.NotificationsDisabledCategories = notifications.DisableCategories(notifications.Deprecation)
 		config.NotificationsDisabledClassifications = notifications.DisableClassifications(notifications.Deprecation)
 
--- a/neo4j/config_test.go
+++ b/neo4j/config_test.go
@@ -126,6 +126,7 @@ func TestValidateAndNormaliseConfig(rt *testing.T) {
 	rt.Run("Configure both NotificationsDisabledCategories and NotificationsDisabledCategories", func(t *testing.T) {
 		config := defaultConfig()
 
+		//lint:ignore SA1019 NotificationsDisabledCategories and DisableCategories are supported at least until 6.0
 		config.NotificationsDisabledCategories = notifications.DisableCategories(notifications.Deprecation)
 		config.NotificationsDisabledClassifications = notifications.DisableClassifications(notifications.Deprecation)
 
--- a/neo4j/db/summary.go
+++ b/neo4j/db/summary.go
@@ -90,6 +90,8 @@ type ProfiledPlan struct {
 	Time              int64
 }
 
+// StreamSummary is part of the GQL compliant notifications preview feature
+// (see README on what it means in terms of support and compatibility guarantees)
 type StreamSummary struct {
 	HadRecord bool
 	HadKey    bool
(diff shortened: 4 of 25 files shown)


StephenCathcart:
[[Go] GQL Notifications Implementation to ADR](https://trello.com/c/DSvToIpB/316-go-gql-notifications-implementation-to-adr)

neo4j/config/driver.go
```diff
@@ -182,10 +182,20 @@ type Config struct {
 	// By default, the server's settings are used.
 	// Disabling severities allows the server to skip analysis for those, which can speed up query execution.
 	NotificationsMinSeverity notifications.NotificationMinimumSeverityLevel
+	// Deprecated: please use NotificationsDisabledClassifications. This will be removed in 6.0.
+	//
 	// NotificationsDisabledCategories defines the categories of notifications the server should not send.
 	// By default, the server's settings are used.
 	// Disabling categories allows the server to skip analysis for those, which can speed up query execution.
+	//
+	//lint:ignore SA1019 NotificationsDisabledCategories is supported at least until 6.0
```
robsdedude: I strongly dislike the fact that this lint ignore comment appears in the docs! But I couldn't find a way to satisfy `staticcheck` without having this comment show up :disappointed: except for disabling this lint for the whole file which I didn't like either.

neo4j/config_test.go
```diff
@@ -18,6 +18,7 @@
 package neo4j
 
 import (
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/notifications"
 	"math"
 	"testing"
 	"time"
```
robsdedude, line 24:
```suggestion
	"math"
	"testing"
	"time"
	
	"github.com/neo4j/neo4j-go-driver/v5/neo4j/notifications"
```
Not sure if you want to work towards or away from it, but some files have the internal imports separated by a blank line.

N.B., seems the majority of files doesn't have them separated. So maybe that's the easier direction to unify towards.

neo4j/db/summary.go
```diff
@@ -90,6 +90,11 @@ type ProfiledPlan struct {
 	Time              int64
 }
 
+type StreamSummary struct {
```
robsdedude, line 95: What speaks against moving this to `neo4j/internal`? It feels like this in an implementation detail that shouldn't be exposed to users. Especially not undocumented like this.

neo4j/db/summary.go
```diff
@@ -99,6 +104,20 @@ type Notification struct {
 	Category    string
 }
 
+type GqlStatusObject struct {
```
robsdedude, line 115: We should probably add a doc comment to the struct an its fields.

neo4j/db/summary.go
```diff
@@ -99,6 +104,20 @@ type Notification struct {
...
+	// Deprecated: for backward compatibility with Notification.Code only.
+	Code string
+	// Deprecated: for backward compatibility with Notification.Title only.
+	Title             string
```
robsdedude:
```suggestion
```
If I recall the ADR correctly, these fields shouldn't exist here. Users need to check the `.Notifications` if they want that info. Please double-check, though. Feels weird to introduce something that's deprecated from the get-go.

neo4j/db/summary.go
```diff
@@ -127,7 +146,9 @@ type Summary struct {
 	Plan                  *Plan
 	ProfiledPlan          *ProfiledPlan
 	Notifications         []Notification
+	GqlStatusObjects      []GqlStatusObject
 	Database              string
 	ContainsSystemUpdates *bool
 	ContainsUpdates       *bool
+	StreamSummary         StreamSummary
```
robsdedude, line 190: Again: should probably not be public.

neo4j/internal/bolt/bolt3.go
```diff
@@ -652,6 +652,7 @@ func (b *bolt3) receiveNext(ctx context.Context) (*db.Record, *db.Summary, error
 		sum.Minor = b.minor
 		sum.ServerName = b.serverName
 		sum.TFirst = b.currStream.tfirst
+		sum.StreamSummary = b.currStream.ToSummary()
```
robsdedude, line 656: If I interpret the code right, `b.currStream.HadRecord` will *always* be `false`.

robsdedude: There are also no tests that cover the whole `ToSummary()`/`currStream` management dance.

neo4j/internal/db/connection.go
```diff
@@ -59,24 +59,32 @@ type TxConfig struct {
...
 	if n.MinSev != notifications.DefaultLevel {
 		meta["notifications_minimum_severity"] = string(n.MinSev)
 	}
-	if n.DisCats.DisablesNone() {
-		meta["notifications_disabled_categories"] = make([]string, 0)
+
+	disabledKey := "notifications_disabled_categories"
+	if minor >= 5 {
+		disabledKey = "notifications_disabled_classifications"
+	}
```
robsdedude: While this seems to be currently correct, it's also asking for bugs later :grimacing:

This only works as long as we only ever use `NotificationConfig::ToMeta(...)` for Bolt6. I think this method should also care about the major protocol version.

neo4j/notifications/notifications.go
```diff
@@ -30,42 +39,199 @@ const (
...
 	return NotificationDisabledCategories{nil, true}
 }
 
+// DisableNoClassifications creates a NotificationDisabledClassifications that enables all classifications.
+// Can be used for NotificationsDisabledClassifications of neo4j.Config and neo4j.SessionConfig.
+func DisableNoClassifications() NotificationDisabledClassifications {
+	return NotificationDisabledClassifications{nil, true}
+}
+
+// Deprecated: please use NotificationDisabledClassifications.DisablesNone. This will be removed in 6.0.
```
robsdedude: Why bother to deprecate the method when the whole struct is deprecated?

Same for the other method(s), won't comment again.

neo4j/notifications/notifications.go
```diff
@@ -30,42 +39,199 @@ const (
...
+		Position:          notification.Position,
+		Classification:    notification.Category,
+		Severity:          notification.Severity,
+		DiagnosticRecord:  diagnosticRecord,
+		IsNotification:    true,
+	}
+}
+
+// ToGqlStatusObjectFromSummary creates a new db.GqlStatusObject based on the context of the db.StreamSummary.
+func ToGqlStatusObjectFromSummary(summary db.StreamSummary) *db.GqlStatusObject {
```
robsdedude: Should this really be exported? Feels very much like an internal implementation detail to me. Same goes for `ToGqlStatusObject`

neo4j/internal/bolt/hydrator.go
```diff
@@ -49,10 +49,12 @@ type success struct {
 	plan               *db.Plan
 	profile            *db.ProfiledPlan
 	notifications      []db.Notification
+	statuses           []db.GqlStatusObject
 	routingTable       *idb.RoutingTable
 	num                uint32
 	configurationHints map[string]any
 	patches            []string
+	streamSummary      db.StreamSummary
```
robsdedude: Either I'm just unable to find it or this field is never actually written to. :wastebasket:?

neo4j/internal/bolt/hydrator.go
```diff
@@ -878,6 +885,20 @@ func parseNotifications(notificationsx []any) []db.Notification {
 	return notifications
 }
 
+func parseGqlStatusObjects(statuses []any) []db.GqlStatusObject {
+	var gqlStatusObjects []db.GqlStatusObject
+	if statuses != nil {
```
robsdedude, line 888: What's this guard good for? Wouldn't this work just fine with `statuses == nil`? On a tangent: this function is never called with `nil`, but relying only on that is maybe not ideal.

neo4j/internal/bolt/hydrator.go
```diff
@@ -952,28 +973,87 @@ func parseProfile(profilex map[string]any) *db.ProfiledPlan {
...
+	return pos
+}
+
 func parseNotification(m map[string]any) db.Notification {
 	n := db.Notification{}
 	n.Code, _ = m["code"].(string)
-	n.Description = m["description"].(string)
+	if description, found := m["description"].(string); found {
+		n.Description = description
+	}
```
robsdedude:
```suggestion
	if description, ok := m["description"].(string); ok {
		n.Description = description
	}
```

neo4j/internal/bolt/hydrator.go
```diff
@@ -952,28 +973,87 @@ func parseProfile(profilex map[string]any) *db.ProfiledPlan {
...
-	if exists {
-		pos := &db.InputPosition{}
-		i, _ := posx["column"].(int64)
-		pos.Column = int(i)
-		i, _ = posx["line"].(int64)
-		pos.Line = int(i)
-		i, _ = posx["offset"].(int64)
-		pos.Offset = int(i)
-		n.Position = pos
+	if pos, exists := m["position"].(map[string]any); exists {
```
robsdedude:
```suggestion
	if pos, ok := m["position"].(map[string]any); ok {
```

neo4j/internal/bolt/hydrator.go
robsdedude: Nit: `found` and then `exists` further down. Consistency for the ~OCD~ _win_.

neo4j/internal/bolt/hydrator.go
```diff
@@ -952,28 +973,87 @@ func parseProfile(profilex map[string]any) *db.ProfiledPlan {
...
 
+func newDefaultDiagnosticRecord() map[string]any {
+	return map[string]any{
+		"OPERATION":      "",
+		"OPERATION_CODE": "0",
+		"CURRENT_SCHEMA": "/",
+	}
+}
+
+func parseGqlStatusObject(m map[string]any) db.GqlStatusObject {
```
robsdedude, line 1014: Why is there inconsistency in how types from the server are cast? Some fields are cast using panicking casts, others are lenient.

robsdedude, line 1014: Nit: inconsistent naming `found` vd `ok`.

neo4j/internal/bolt/hydrator.go
```diff
@@ -952,28 +973,87 @@ func parseProfile(profilex map[string]any) *db.ProfiledPlan {
...
+	}
+
+	// Backward compatibility support for deprecated Notification API.
+	if title, found := m["title"].(string); found {
+		//lint:ignore SA1019 Title is supported at least until 6.0
+		g.Title = title
+	}
+
+	// Initialize the default diagnostic record
+	diagnosticRecord := newDefaultDiagnosticRecord()
```
robsdedude, line 1045: This is good enough for now, I just want us to be aware that the defaults might change in a future bolt version meaning at that point meaning the could would then need refactoring for this bit to be bolt-version-aware.

neo4j/resultsummary.go
```diff
@@ -228,13 +247,73 @@ type Notification interface {
...
+	Category() notifications.NotificationCategory
+}
+
+// GqlStatusObject represents a GqlStatusObject generated when executing a statement.
+// A GqlStatusObject can be visualized in a client pinpointing problems or other information about the statement.
+// Contrary to failures or errors, GqlStatusObjects do not affect the execution of the statement.
+//
+// GqlStatusObject is part of the GQL compliant notifications preview feature
+// (see README on what it means in terms of support and compatibility guarantees)
+type GqlStatusObject interface {
```
robsdedude, line 264: I just realized that this PR introduces 2 public `GqlStatusObject`. The struct in `neo4j/db` and this interface in `neo4j`. Is there a good reason for this? Could we do with only one of them?

neo4j/internal/bolt/hydrator.go
```diff
@@ -952,28 +973,87 @@ func parseProfile(profilex map[string]any) *db.ProfiledPlan {
...
+	pos := &db.InputPosition{}
+	if i, ok := m["column"].(int64); ok {
+		pos.Column = int(i)
+	}
+	if i, ok := m["line"].(int64); ok {
+		pos.Line = int(i)
+	}
+	if i, ok := m["offset"].(int64); ok {
+		pos.Offset = int(i)
+	}
```
robsdedude, line 987:
```suggestion
	if i, ok := m["column"].(int64); ok {
		pos.Column = i
	}
	if i, ok := m["line"].(int64); ok {
		pos.Line = i
	}
	if i, ok := m["offset"].(int64); ok {
		pos.Offset = i
	}
```
Why is this a lossy cast? Shouldn't `InputPosition.Offset` etc. be int64 instead?

neo4j/internal/bolt/notifications.go
```diff
@@ -28,7 +28,9 @@ func checkNotificationFiltering(
 	bolt idb.Connection,
 ) error {
 	if notificationConfig.MinSev == notifications.DefaultLevel &&
-		!notificationConfig.DisCats.DisablesNone() && len(notificationConfig.DisCats.DisabledCategories()) == 0 {
+		//lint:ignore SA1019 DisablesNone and DisabledCategories are supported at least until 6.0
+		!notificationConfig.DisCats.DisablesNone() && len(notificationConfig.DisCats.DisabledCategories()) == 0 &&
+		!notificationConfig.DisClas.DisablesNone() && len(notificationConfig.DisClas.DisabledClassifications()) == 0 {
```
robsdedude: I think we could simplify the driver code a fair bit if we can at some higher level in the driver unify the two slices inside of `NotificationsDisabledCategories` and `NotificationsDisabledClassifications`. After giving it only a short thought, I'm unsure how to approach this without altering the public API, which I'd rather not.

neo4j/notifications/notifications.go
```diff
@@ -30,42 +39,199 @@ const (
...
+	}
+}
+
+// ToGqlStatusObject returns a db.GqlStatusObject that corresponds to the given db.Notification.
+// It maps fields from the notification to their respective status fields.
+func ToGqlStatusObject(notification db.Notification) *db.GqlStatusObject {
+	defaultStatus := newUnknownInformationResultGqlStatusObject()
+	if notification.Severity == string(WarningLevel) {
+		defaultStatus = newUnknownWarningResultGqlStatusObject()
+	}
```
robsdedude:
```suggestion
	var defaultStatus *db.GqlStatusObject
	if notification.Severity == string(WarningLevel) {
		defaultStatus = newUnknownWarningResultGqlStatusObject()
	} else {
		defaultStatus = newUnknownInformationResultGqlStatusObject()
	}
```
Avoid potentially wasted allocations.

neo4j/notifications/notifications.go
```diff
@@ -30,42 +39,199 @@ const (
...
+		StatusDescription: "info: unknown notification",
+		DiagnosticRecord:  newDefaultDiagnosticRecord(),
+	}
+}
+
+// ToGqlStatusObject returns a db.GqlStatusObject that corresponds to the given db.Notification.
+// It maps fields from the notification to their respective status fields.
+func ToGqlStatusObject(notification db.Notification) *db.GqlStatusObject {
+	defaultStatus := newUnknownInformationResultGqlStatusObject()
+	if notification.Severity == string(WarningLevel) {
```
robsdedude:
```suggestion
	if notification.Severity == string(Warning) {
```
Nit: not that this would make any difference :sweat_smile: it's just semantically more precise as we're not dealing with a severity filter here.

neo4j/config.go
robsdedude, line 1: This comment is a general comment and does not belong to this file. I just don't like how general GitHub comments can't be resolved and don't support threads.

Not every new public API surface introduced in this PR is marked as preview feature. They should unless we intend to keep them around independently of this feature.

neo4j/resultsummary.go
```diff
@@ -504,14 +599,69 @@ func (p *profile) Time() int64 {
...
-		notifications[i] = &notification{notification: &s.sum.Notifications[i]}
+
+	if s.sum.GqlStatusObjects != nil {
+		// Polyfill Notifications from GqlStatusObjects
+		n := make([]Notification, 0, len(s.sum.GqlStatusObjects))
+		for i := range s.sum.GqlStatusObjects {
+			status := s.sum.GqlStatusObjects[i]
+			if status.IsNotification {
+				n = append(n, &notification{notification: notifications.ToNotification(status)})
+
```
robsdedude:
```suggestion
```

neo4j/resultsummary.go
```diff
@@ -586,9 +737,96 @@ func (n *notification) Position() InputPosition {
...
+func (g *gqlStatusObject) StatusDescription() string {
+	return g.gqlStatusObject.StatusDescription
+}
+
+func (g *gqlStatusObject) Position() InputPosition {
+	if g.gqlStatusObject.Position == nil {
+		return nil
+	}
+	return g
+}
```
robsdedude, line 760: Is there any good reason to have this method, if all it does is return itself? This also means this `g.Offset()` et al. might blow up when called while `g.gqlStatusObject.Position == nil` or am I wrong?

neo4j/notifications/notifications.go
```diff
@@ -17,8 +17,17 @@
 
 package notifications
 
+import (
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/db"
+)
+
+// Deprecated: please use NotificationClassification. This will be removed in 6.0.
```
robsdedude: Not yet to be deprecated as the alternative is still in preview.

Actually, this holds true for more items. Basically everything that's going to be replaced by some GQL counterpart has to stay undeprecated until that counterpart is GA and can be fully relied upon by users.

neo4j/notifications/notifications.go
```diff
@@ -30,42 +39,199 @@ const (
...
+	statusDescription := notification.Description
+	if statusDescription == "" {
+		statusDescription = defaultStatus.StatusDescription
+	}
+
+	diagnosticRecord := newDefaultDiagnosticRecord()
+
+	if notification.Position != nil {
+		diagnosticRecord["_position"] = notification.Position
+	}
```
robsdedude: The diagnostic record is supposed to be a raw map as it would come from the server. There shouldn't be a parsed `Position` struct in there.

testkit-backend/2cypher.go
```diff
@@ -201,7 +202,17 @@ func nativeToCypher(v any) map[string]any {
...
+	case *db.InputPosition:
+		if x != nil {
+			return valueResponse("CypherMap", map[string]any{
+				"column": nativeToCypher(int64(x.Column)),
+				"line":   nativeToCypher(int64(x.Line)),
+				"offset": nativeToCypher(int64(x.Offset)),
+			})
+		}
+		return nil
 	}
```
robsdedude: This should never happen. The server cannot send this type and the driver shouldn't put this type into anything that's supposed to polyfill for things that could come from the server.

testkit-backend/backend.go
```diff
@@ -1440,8 +1446,11 @@ func serializeRecord(record *neo4j.Record) map[string]any {
 	return data
 }
 
-func serializeNotifications(slice []neo4j.Notification) []map[string]any {
+func serializeNotifications(slice []neo4j.Notification, version db.ProtocolVersion) []map[string]any {
 	if slice == nil {
+		if version.Minor >= 5 {
+			return []map[string]any{}
+		}
```
robsdedude: I think we've had a short chat about it, but I can't remember: could you please explain to me why this special casing for some bolt version is required?

Also, I'm almost certain that this should not only depend on the protocol's minor version, but also or only on the major version.

testkit-backend/backend.go
```diff
@@ -1471,6 +1480,45 @@ func serializeNotifications(slice []neo4j.Notification) []map[string]any {
 	return res
 }
 
+func serializeGqlStatusObjects(slice []neo4j.GqlStatusObject) []map[string]any {
+	if slice == nil {
+		return []map[string]any{}
+	}
+	if len(slice) == 0 {
+		return []map[string]any{}
+	}
```
robsdedude: The first is a special case of the second. So only the second would be required, if I read this right.

Further, doesn't `range` over a `nil` map simply execute 0 times making the whole function just work for `nil` without any guards at the top?

robsdedude (changes requested):

neo4j/config_test.go
StephenCathcart: I did a little check myself and you're right, most of the files skip the blank line. I'll work towards that to unify the files and see about updating my linting to do this automatically 👍

neo4j/notifications/notifications.go
StephenCathcart: Understood, I've removed deprecation warnings from most of the older notification stuff until status objects are out of Preview.

neo4j/config/driver.go
StephenCathcart: lint is now removed due to removing deprecation due to gql changes being in preview

testkit-backend/backend.go
StephenCathcart: The conversation was pretty much:

> Before 5.5, when the server doesn't send notifications, our drivers behave differently. Some (I think most) return an empty list and they skip this test. Starting with 5.5, the server never sends notifications, but the driver is supposed to polyfill notifications from statuses. And if there are no notifications in the statuses, the driver should return an empty list. TL;DR: unspecified behavior before 5.5.

What was happening was that the Go driver returned nil when serializing notifications, not an empty slice. Now that we expect an empty slice after bolt 5.5, the Testkit tests also expect an empty slice. So this was to prevent older tests from breaking. You're right about the major version though, I've updated the check to include major 4 too.

testkit-backend/backend.go
StephenCathcart: Yep think I added the guard before I refactored the function to return an empty map by default so will work fine without the guards. Updated

neo4j/notifications/notifications.go
StephenCathcart: deprecations have been removed due to GQL status objects being in preview.

neo4j/internal/bolt/hydrator.go
StephenCathcart: It might work fine without the guard given how h.array() works. I implemented it like this to match the `parseNotifications` implementation: https://github.com/StephenCathcart/neo4j-go-driver/blob/016d5f1676960df0ced1b93defc328b6042e50e9/neo4j/internal/bolt/hydrator.go#L874-L876.

StephenCathcart: TestKit tests still pass with the guard removed, happy to keep or remove the guard.

neo4j/internal/bolt/hydrator.go
StephenCathcart: I've updated this to `ok` to match the rest of the file, along with others.

neo4j/internal/bolt/hydrator.go
StephenCathcart: resolved in conversation above

neo4j/internal/bolt/hydrator.go
StephenCathcart: resolved in above conversation.

neo4j/internal/bolt/hydrator.go
StephenCathcart: Honestly no idea this was refactored away from the parseNotification function/original code we had for reuse in the parseGqlStatusObject function, I assumed there was some hidden logic in why it was implemented way in the first place.

neo4j/internal/bolt/hydrator.go
StephenCathcart: Pretty much just to be consistent with parseNotifications, however, you're right I've updated them to all be more lenient, and tests are still fine.

neo4j/internal/bolt/hydrator.go
StephenCathcart: Good to know thanks I'll keep this in mind.

neo4j/notifications/notifications.go
StephenCathcart: Good spot

neo4j/resultsummary.go
StephenCathcart: It returns itself which is an interface of InputPosition, so from the users code they will write:

```
for _, gqlStatusObject := range gqlStatusObjects {
    pos := gqlStatusObject.Position()
    // you can't do gqlStatusObject.Offset() as the interface doesn't allow it. You're expected to check if
    // gqlStatusObject.Position() is nil before accessing gqlStatusObject.Position().Line() etc.
}
```

It's the same design pattern as the Notification interface: https://github.com/neo4j/neo4j-go-driver/blob/2525e8f57093834960c55afd7519869df8954bfc/neo4j/resultsummary.go#L729-L746

Happy to discuss

neo4j/notifications/notifications.go
StephenCathcart: Yep good spot this is why I needed to update `nativeToCypher` in the backend to handle InputPosition which I've also now removed.

testkit-backend/2cypher.go
StephenCathcart: See above conversation, this was happening due to passing InputPosition to diagnostic record rather than a raw map, this is now removed.

neo4j/internal/db/connection.go
StephenCathcart: I've updated this to take in `version db.ProtocolVersion` instead which has access to both major and minor version.

neo4j/internal/bolt/notifications.go
StephenCathcart: Had a quick thought myself, couldn't think of a clean way, can have a chat about this.

neo4j/internal/bolt/hydrator.go
StephenCathcart: One thing to think about is that the InputPosition interface exposes all ints, not int64, so this would be a breaking change for the older notification stuff.

neo4j/config.go
StephenCathcart: Think I've added preview warnings to all relevant APIs 👍

neo4j/db/summary.go
StephenCathcart: The `GqlStatusObject` interface that's exposed to users with `result.Summary.GqlStatusObjects()` doesn't have `Code` or `Title` so these won't be shown to the user, they're included in the struct to be able to store both Code and Title in the hydrator when parsing from the server:

https://github.com/neo4j/neo4j-go-driver/blob/016d5f1676960df0ced1b93defc328b6042e50e9/neo4j/internal/bolt/hydrator.go#L1024

then the values are read and written to in notifications.go when converting between Notifications and GqlStatusObjects, if I make these fields private I'll need to think of how else we can store this data between the packages.

neo4j/db/summary.go
StephenCathcart: Yeah, there seems to be some inconsistency in the summary.go file, things like Notification and fields have no comments (which I copied the pattern from), others like ProfiledPlan have field comments but none for the later fields that were added at a later date. I'll add comments to this struct now and raise an issue to fix the other comments at a later date.

neo4j/db/summary.go
StephenCathcart: leaving open for discussion - see below conversation around the struct/interface discussion in the driver

neo4j/internal/bolt/hydrator.go
StephenCathcart: leaving open for discussion

neo4j/resultsummary.go
StephenCathcart: Pretty much just the pattern I followed for most of the Go driver API:

`Notification` struct in summary.go -> `Notification` interface in resultsummary.go
`Plan` struct in summary.go -> `Plan` interface in resultsummary.go
`ProfiledPlan` struct in summary.go -> `ProfiledPlan` interface in resultsummary.go
`InputPosition` struct in summary.go -> `InputPosition` interface in resultsummary.go

etc etc.

I think it is done like this as most of the driver API surface returns interfaces rather than raw structs to let us have some flexibility on updating the implementation details of the structs without breaking the exposed interfaces (this ties in with the above conversation of having Code and Title in the GqlStatusObject struct and not the GqlStatusObject interface).

StephenCathcart: leaving open for discussion

neo4j/notifications/notifications.go
StephenCathcart: Agreed, I wasn't sure how to handle it but I've created an internal notification package and moved `ToNotification`, `ToGqlStatusObject` and `ToGqlStatusObjectFromSummary` to this along with all static GqlStatusObjects.

neo4j/internal/bolt/hydrator.go
StephenCathcart: correct this was the old way of doing it before I added `ToSummary` to `stream` itself.

neo4j/internal/bolt/bolt3.go
StephenCathcart: Correct, in bolt4/5 we have `pullResponseHandler` where we set hadRecord to true if the record is not nil in onRecord. Unsure on where to set this in bolt3, be good to pair on this.

P.S will add tests to the dance

neo4j/internal/bolt/hydrator.go
StephenCathcart: Packstream deals with int64 - this is an abstraction, other drivers do this.

neo4j/resultsummary.go
StephenCathcart: As discussed this morning, ideally, we might be able to move most of these public structs to the internal package and only export the interface counterparts. Currently gives lots of cyclic dependency issues/exposes internal structs to the client. Needs more thought I've carded this up for future improvement to driver.

neo4j/internal/bolt/hydrator.go
StephenCathcart: leaving to be consistent with notification parsing logic.

neo4j/db/summary.go
StephenCathcart: As discussed, not ideal to have fields that are for internal use exposed on a public struct, but most of the driver seems to be done this way (struct with an equivalent interface which is then returned on most of the driver API). Similar to the conversation below, we need a future discussion to perhaps move these structs to an internal package to stop these things from happening.

neo4j/internal/bolt/bolt3.go
StephenCathcart: bolt3 hadRecord has been fixed

neo4j/db/summary.go
StephenCathcart: See below conversations below around other structs design in the driver -  might be able to move most of these public structs to the internal package and only export the interface counterparts.

neo4j/db/summary.go
StephenCathcart: I've tried to refactor this to an internal package which would mean that the StreamSummary that's attached to the Summary struct would be exposing an internal struct which isn't ideal. Think I'll wrap this up with a future refactor of all the public structs that we have

StephenCathcart:
[[Go] Add `description` to `statuses`](https://trello.com/c/uL5R8pkx/375-go-add-description-to-statuses)

neo4j/internal/bolt/bolt3.go
StephenCathcart: tests covering to summary have been added to bolt3-5 tests

Issue #590, opened 2024-07-24
panic: runtime error: makeslice: len out of range


mindaugasrukas:
We are getting panics with the message: `runtime error: makeslice: len out of range`. More details below.

**Neo4j Version:** 5 (neo4j Aura)
**Neo4j Mode**: Single instance  
**Driver version**: Go driver v5.22.0
**Operating System:** alpine:3.20 (Docker on K8s)

### Steps to reproduce
Unknown yet

We assume the issue is related to the length of the path of a result. We don't see this issue on small paths.

CYPHER query example we execute:
```
MATCH (s1:Item {x_uuid: $x_uuid, eid: $eid1})
WITH s1
MATCH (s2:Item {x_uuid: $x_uuid, eid: $eid2})
WITH s1, s2
MATCH p = shortestPath((s1)-[r:NEXT*0..500]->(s2)),
	p1 = (before:Item)-[:NEXT*0..500]->(s1),
	p2 = (s2)-[:NEXT*0..500]->(after:Item),
	path = shortestPath((before)-[:NEXT*0..500]->(after))
WHERE
	NONE(n IN nodes(p1)[1..-1] WHERE n:Item)
	AND NONE(n IN nodes(p2)[1..-1] WHERE n:Item)
RETURN path
```

Explanation of the query:

1. We have a graph of connected, call it, Items and NotItems. For example: (:Item)-[:NEXT]->(:NotItem)-[:NEXT]->(:NotItem)-[:NEXT]->(:Item)-[:NEXT]->(:Item)-[:NEXT]->(:NotItem)-[:NEXT]->(:NotItem)-[:NEXT]->(:Item)-...
2. we find the shortest path between s1 and s2 Items.
3. then we expand the path by adding the closest Items at the beginning and the end (if they exist).
4. the final `path = shortestPath(...)` is probably not needed. This is just to make it easier to parse the results later. We could return `RETURN p1, p, p2`, but that makes it complicated to build the full end-to-end path.

So far, we observe that if a final path is somewhere more than 100 nodes long (not sure exactly), we get a panic. The largest of our paths might be around 1000 nodes long.

We have this index:
```
CREATE CONSTRAINT FOR (n:Item) REQUIRE (n.eid, n.x_uuid) IS UNIQUE;
```

### Expected behavior
no error

### Actual behavior

The client failed to run the query with a panic `runtime error: makeslice: len out of range`. The stacktrace is attached below.

Stack trace (I removed unrelated internal library names):
[...]

PR #592, opened 2024-08-09
Fix hydrator path overflow bug

StephenCathcart:
Fix for https://github.com/neo4j/neo4j-go-driver/issues/590. A new test case was added for regression.

### DIFF
--- a/neo4j/internal/bolt/hydrator_test.go
+++ b/neo4j/internal/bolt/hydrator_test.go
@@ -19,6 +19,7 @@ package bolt
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"runtime/debug"
 	"testing"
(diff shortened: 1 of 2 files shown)


neo4j/internal/bolt/hydrator_test.go
```diff
@@ -1067,6 +1069,72 @@ func TestHydratorBolt5(outer *testing.T) {
...
+			}
+			// Indices
+			packer.ArrayHeader(c.size)
+			for i := 1; i <= c.size; i++ {
+				packer.Int(1)
+			}
+			// Test we can hydrate without a panic from int overflows.
+			buf, err := packer.End()
+			if err != nil {
+				panic("Build error")
```
fbiville: nit: maybe `t.Fatal` instead of `panic`? (same below)

bigmontz: What is the difference for the user experience?

bigmontz: (now I see it is test code 🤦🏼 )

bigmontz (approved):
💯

neo4j/internal/bolt/hydrator_test.go
StephenCathcart: agreed thanks for the feedback 👍

PR #598, opened 2024-08-28
add Records method to ResultWithContext

jordan-bonecutter:
With the introduction of go 1.23 it would be useful to add support for an `iter.Seq` from the `ResultWithContext` type to avoid allocating memory on the order of the size of the results. This PR adds a `Records` method returning an `iter.Seq2[*Record, error]` which allows the user to for loop over the records in a `ResultWithContext`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
 module github.com/neo4j/neo4j-go-driver/v5
 
-go 1.18
+go 1.23.0
--- a/neo4j/driver_with_context_test.go
+++ b/neo4j/driver_with_context_test.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	. "github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/testutil"
+	"iter"
 	"net/url"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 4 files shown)


StephenCathcart:
Thanks for your contribution @jordan-bonecutter! Unfortunately, for backward compatibility reasons, the Neo4j Go driver only bumps the Go version (currently 1.18) during a major release. Implementing this change now would break backward compatibility, so we would need to hold off until version 6 of the drivers (we're currently on version 5.24.0).

Please feel free to keep an eye on our progress towards version 6, I've carded this up to revisit at that time.

jordan-bonecutter:
> Thanks for your contribution @jordan-bonecutter! Unfortunately, for backward compatibility reasons, the Neo4j Go driver only bumps the Go version (currently 1.18) during a major release. Implementing this change now would break backward compatibility, so we would need to hold off until version 6 of the drivers (we're currently on version 5.24.0).
> 
> Please feel free to keep an eye on our progress towards version 6, I've carded this up to revisit at that time.

Makes sense.

What if instead of returning `iter.Seq2[*db.Record, error]` it returned a `func(yield func(*db.Record, error) bool)` which could be inferred to be an `iter.Seq2` by newer versions of go without requiring the `1.23` bump? I can prototype this but it should work.

jordan-bonecutter:
@StephenCathcart I pushed a new commit which doesn't require go `1.23` in the `go.mod` file but works as an `iter.Seq2`. Let me know if you're ok with opening another PR

PR #601, opened 2024-09-25
Bugfix: Retry mechanism doesn't respect the context

SeBBBe:
This PR removes the `Sleep` func in the retry state and instead uses the following contruct to either wait the determined amount of time, _or_ cancel if the context is canceled. I also added a test case that reproduces this scenario.

```
select {
case <-ctx.Done():
	s.Errs = []error{context.Canceled}
	return false
case <-time.After(sleepTime):
}
```

### DIFF
--- a/neo4j/internal/retry/state.go
+++ b/neo4j/internal/retry/state.go
@@ -37,7 +37,6 @@ type State struct {
 	Log                     log.Logger
 	LogName                 string
 	LogId                   string
-	Sleep                   func(time.Duration)
 	Throttle                Throttler
 	MaxDeadConnections      int
 	DatabaseName            string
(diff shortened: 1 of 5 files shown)


neo4j/internal/retry/state.go
```diff
@@ -103,7 +102,13 @@ func (s *State) Continue() bool {
 		sleepTime := s.Throttle.delay()
 		s.Log.Debugf(s.LogName, s.LogId,
 			"Retrying transaction (%s): %s [after %s]", s.cause, lastErr, sleepTime)
-		s.Sleep(sleepTime)
+
+		select {
+		case <-ctx.Done():
+			s.Errs = []error{context.Canceled}
```
StephenCathcart:
```suggestion
			s.Errs = append(s.Errs, context.Canceled)
```

StephenCathcart: Potential loss of errors as we discussed.

robsdedude (approved):
:timer_clock: :zap: :stop_sign:

StephenCathcart (approved):
💯

PR #607, opened 2024-10-09
ADR 023: GQL Compliant Errors

StephenCathcart:
⚠️ This feature is in PREVIEW and may change without following the deprecation policy.
 
This PR adds GQL Error support to the Neo4jError struct. The new fields are added in a backward-compatible way, meaning existing error messages and codes will remain functional, even for GQL-compliant errors. Additionally, this update includes a polyfill function to set default values for GQL fields if they are not provided.

This is part of a broader effort to align Neo4j drivers with GQL standards.

### DIFF
--- a/neo4j/db/errors.go
+++ b/neo4j/db/errors.go
@@ -32,22 +32,51 @@ const (
 	UnknownError   ErrorClassification = "UNKNOWN"
 )
 
-// Neo4jError is created when the database server failed to fulfill request.
-// TODO 6.0: rename to GqlError
+// TODO: when we no longer need to support the old Neo4jError, rename the following fields:
+// - Neo4jError -> GqlError
+// - GqlStatusDescription -> StatusDescription
+// - GqlClassification -> Classification
+// - GqlRawClassification -> RawClassification
+// - GqlDiagnosticRecord -> DiagnosticRecord
+// - GqlCause -> Cause
+// - GqlStatus to remain as-is to comply with GQLSTATUS.
+
+// Neo4jError is created when the database server fails to fulfill a request.
 type Neo4jError struct {
-	// TODO 6.0: to be deprecated.
-	Code      string
-	Msg       string
+	// Code is the Neo4j-specific error code, to be deprecated in favor of GqlStatus.
+	Code string
+	// Msg is the specific error message describing the failure.
+	Msg string
+	// GqlStatus returns the GQLSTATUS.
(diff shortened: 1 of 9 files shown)


StephenCathcart:
[[Go] Implement GQL Errors as per ADR](https://trello.com/c/1ufoO1b8/472-go-implement-gql-errors-as-per-adr)

neo4j/db/errors.go
```diff
@@ -23,12 +23,34 @@ import (
...
+
+const (
+	ClientError    ErrorClassification = "CLIENT_ERROR"
+	DatabaseError  ErrorClassification = "DATABASE_ERROR"
+	TransientError ErrorClassification = "TRANSIENT_ERROR"
+	UnknownError   ErrorClassification = "UNKNOWN"
+)
+
 // Neo4jError is created when the database server failed to fulfill request.
+// TODO 6.0: rename to GqlError
```
robsdedude: As discussed on the call, reconsider if this (and the other to dos for 6.0)  should remain. Maybe only some, maybe all, maybe none.

If you decide to keep some/all, please be aware that these comments show up in the API docs. :grimacing: not so nice.

neo4j/db/errors.go
```diff
@@ -23,12 +23,34 @@ import (
...
+	// TODO 6.0: rename to StatusDescription
+	GqlStatusDescription string
+	// TODO 6.0: rename to Classification
+	GqlClassification ErrorClassification
+	// TODO 6.0: rename to RawClassification
+	GqlRawClassification string
+	// TODO 6.0: rename to DiagnosticRecord
+	GqlDiagnosticRecord map[string]any
+	// TODO 6.0: rename to Cause
+	GqlCause       *Neo4jError
```
robsdedude: Gql fields are supposed to go out in preview.

robsdedude: Some docs for those fields would be nice. Have a look at the ADR and/or other drivers for inspiration.

robsdedude (commented):
Looking very good so far :sparkles:

neo4j/db/errors.go
StephenCathcart: I moved the TODOs regarding future renaming above the structure so as not to interfere with API docs.

neo4j/db/errors.go
StephenCathcart: Added preview comments.

StephenCathcart: Docs added 👍

PR #617, opened 2025-01-26
Optimize serialization of []any and map[string]any

zolstein:
Improve serialization performance by specializing slices or maps of `any`, and optimizing reflective looping through maps.

Special-casing `[]any` and especially `map[string]any`, which are common parameter types, eliminates allocation overhead to loop through these values.

Using `MapRange` rather than `MapKeys` and `MapIndex` (and using `Value.SetIterKey/Value` rather than `MapIter.Key/Value`) significantly reduces allocations when looping through other map types.

### DIFF
--- a/neo4j/internal/bolt/outgoing.go
+++ b/neo4j/internal/bolt/outgoing.go
@@ -408,6 +408,11 @@ func (o *outgoing) packX(x any) {
 			o.packer.Strings(s)
 		case []float64:
 			o.packer.Float64s(s)
+		case []any:
+			o.packer.ArrayHeader(len(s))
+			for _, e := range s {
+				o.packX(e)
+			}
 		default:
 			num := v.Len()
 			o.packer.ArrayHeader(num)
@@ -422,17 +427,27 @@ func (o *outgoing) packX(x any) {
 			o.packer.IntMap(m)
 		case map[string]string:
 			o.packer.StringMap(m)
+		case map[string]any:
+			o.packMap(m)
 		default:
 			t := reflect.TypeOf(x)
 			if t.Key().Kind() != reflect.String {
 				o.onPackErr(&db.UnsupportedTypeError{Type: reflect.TypeOf(x)})
 				return
 			}
-			o.packer.MapHeader(v.Len())
-			// TODO Use MapRange when min Go version is >= 1.12
-			for _, ki := range v.MapKeys() {
-				o.packer.String(ki.String())
-				o.packX(v.MapIndex(ki).Interface())
+			l := v.Len()
+			o.packer.MapHeader(l)
+			if l == 0 {
+				return
+			}
+			key := reflect.New(t.Key()).Elem()
+			value := reflect.New(t.Elem()).Elem()
+			r := v.MapRange()
+			for r.Next() {
+				key.SetIterKey(r)
+				value.SetIterValue(r)
+				o.packer.String(key.String())
+				o.packX(value.Interface())
 			}
 		}
 	default:
--- a/neo4j/internal/bolt/outgoing.go
+++ b/neo4j/internal/bolt/outgoing.go
@@ -391,7 +391,7 @@ func (o *outgoing) packX(x any) {
 		case reflect.Struct:
 			o.packStruct(x)
 		default:
-			o.packX(i.Interface())
+			o.packV(i)
 		}
 	case reflect.Struct:
 		o.packStruct(x)
@@ -417,7 +417,7 @@ func (o *outgoing) packX(x any) {
 			num := v.Len()
 			o.packer.ArrayHeader(num)
 			for i := 0; i < num; i++ {
-				o.packX(v.Index(i).Interface())
+				o.packV(v.Index(i))
 			}
 		}
 	case reflect.Map:
@@ -448,16 +448,50 @@ func (o *outgoing) packX(x any) {
 			}
 			key := reflect.New(t.Key()).Elem()
 			value := reflect.New(t.Elem()).Elem()
-			r := v.MapRange()
-			for r.Next() {
-				key.SetIterKey(r)
-				value.SetIterValue(r)
+			iter := v.MapRange()
+			for iter.Next() {
+				key.SetIterKey(iter)
+				value.SetIterValue(iter)
 				o.packer.String(key.String())
-				o.packX(value.Interface())
+				o.packV(value)
 			}
 		}
 	default:
 		o.onPackErr(&db.UnsupportedTypeError{Type: reflect.TypeOf(x)})
+	}
+}
+
+func (o *outgoing) packV(v reflect.Value) {
+	switch v.Kind() {
+	case reflect.Bool:
+		o.packer.Bool(v.Bool())
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		o.packer.Int64(v.Int())
+	case reflect.Uint8, reflect.Uint16, reflect.Uint32:
+		o.packer.Uint32(uint32(v.Uint()))
+	case reflect.Uint64, reflect.Uint:
+		o.packer.Uint64(v.Uint())
+	case reflect.Float32, reflect.Float64:
+		o.packer.Float64(v.Float())
+	case reflect.String:
+		o.packer.String(v.String())
+	case reflect.Ptr:
+		if v.IsNil() {
+			o.packer.Nil()
+			return
+		}
+		// Inspect what the pointer points to
+		i := reflect.Indirect(v)
+		switch i.Kind() {
+		case reflect.Struct:
+			o.packStruct(v.Interface())
+		default:
+			o.packV(i)
+		}
+	case reflect.Struct:
+		o.packStruct(v.Interface())
+	default:
+		o.packX(v.Interface())
 	}
 }
 
--- a/neo4j/internal/bolt/outgoing.go
+++ b/neo4j/internal/bolt/outgoing.go
@@ -490,6 +490,27 @@ func (o *outgoing) packV(v reflect.Value) {
 		}
 	case reflect.Struct:
 		o.packStruct(v.Interface())
+	case reflect.Slice:
+		switch v.Type().Elem().Kind() {
+		case reflect.Uint8:
+			o.packer.Bytes(v.Bytes())
+			return
+		case reflect.Int, reflect.Int64, reflect.String, reflect.Float64, reflect.Interface:
+			if v.Len() > 5 {
+				// Accept the cost of an allocation (v.Interface())
+				// because this slice type is optimized in packX.
+				// The exact length from which the optimization amortizes
+				// the allocation cost depends on many factors.
+				// 5 is an arbitrary guess.
+				o.packX(v.Interface())
+				return
+			}
+		}
+		num := v.Len()
+		o.packer.ArrayHeader(num)
+		for i := 0; i < num; i++ {
+			o.packV(v.Index(i))
+		}
 	default:
 		o.packX(v.Interface())
 	}
--- a/neo4j/internal/bolt/outgoing.go
+++ b/neo4j/internal/bolt/outgoing.go
@@ -461,6 +461,17 @@ func (o *outgoing) packX(x any) {
 	}
 }
 
+func typeForPrimitive[T any]() reflect.Type {
+	var v T
+	return reflect.TypeOf(v)
+}
+
+var intT = typeForPrimitive[int]()
+var int64T = typeForPrimitive[int64]()
+var stringT = typeForPrimitive[string]()
+var float64T = typeForPrimitive[float64]()
+var anyT = reflect.TypeOf((*any)(nil)).Elem()
+
 func (o *outgoing) packV(v reflect.Value) {
 	switch v.Kind() {
 	case reflect.Bool:
@@ -491,22 +502,26 @@ func (o *outgoing) packV(v reflect.Value) {
 	case reflect.Struct:
 		o.packStruct(v.Interface())
 	case reflect.Slice:
-		switch v.Type().Elem().Kind() {
-		case reflect.Uint8:
+		elemType := v.Type().Elem()
+		if elemType.Kind() == reflect.Uint8 {
 			o.packer.Bytes(v.Bytes())
 			return
-		case reflect.Int, reflect.Int64, reflect.String, reflect.Float64, reflect.Interface:
-			if v.Len() > 5 {
+		}
+		num := v.Len()
+		if num > 5 {
+			switch elemType {
+			case intT, int64T, stringT, float64T, anyT:
 				// Accept the cost of an allocation (v.Interface())
 				// because this slice type is optimized in packX.
 				// The exact length from which the optimization amortizes
 				// the allocation cost depends on many factors.
-				// 5 is an arbitrary guess.
+				// 5 seemed a close to the cross-over point on my system
+				// running Go 1.23.0 linux/amd64.
+				// cf. https://github.com/neo4j/neo4j-go-driver/pull/617
 				o.packX(v.Interface())
 				return
 			}
 		}
-		num := v.Len()
 		o.packer.ArrayHeader(num)
 		for i := 0; i < num; i++ {
 			o.packV(v.Index(i))
--- a/neo4j/internal/bolt/outgoing.go
+++ b/neo4j/internal/bolt/outgoing.go
@@ -423,10 +423,16 @@ func (o *outgoing) packX(x any) {
 	case reflect.Map:
 		// Optimizations
 		switch m := x.(type) {
+		case map[string][]byte:
+			o.packer.BytesMap(m)
 		case map[string]int:
 			o.packer.IntMap(m)
+		case map[string]int64:
+			o.packer.Int64Map(m)
 		case map[string]string:
 			o.packer.StringMap(m)
+		case map[string]float64:
+			o.packer.Float64Map(m)
 		case map[string]any:
 			o.packMap(m)
 		default:
--- a/neo4j/internal/bolt/outgoing_test.go
+++ b/neo4j/internal/bolt/outgoing_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


robsdedude:
Hello, thank you for taking the time to contribute to a Neo4j project, we appreciate all community engagement. 
Before we review a change we require that users have signed our Contributor License Agreement(CLA).
For more details on signing our CLA please see: https://neo4j.com/developer/cla/

zolstein:
@robsdedude I did send an email agreeing to the license last month. Is there some reason that that didn't go through? Is it associated with the wrong email address? In any case, I've sent another agreement associated with a different email address, hopefully that resolves the issue.

robsdedude:
One might think our CLA processes is automated... Well it's only partially 😅 Thanks, I'll look into performing the steps to get CI to pass and review your PR in the next days.

robsdedude:
FYI, I've been having a look and the speed-up looks very promising! I've not spotted any logic issues either :muscle: great job! But you've piqued my curiosity. I'm currently looking into whether there's a way to generalize this optimization so that it applies to even more cases.

robsdedude:
I didn't succeed :/ everything I tried didn't improve the situation or made it even worse. What I've don, however, is added all `map[String]xyz` short paths that also exist for `[]xyz`.

Here are some benchmarks I quickly threw together to validate that this PR makes things faster:

```
### BEFORE ###

BenchmarkOutgoing
BenchmarkOutgoing/pack_[]any
BenchmarkOutgoing/pack_[]any-16                                      	 4242537	       250.7 ns/op
BenchmarkOutgoing/pack_[]bool
BenchmarkOutgoing/pack_[]bool-16                                     	 2663608	       455.2 ns/op
BenchmarkOutgoing/pack_map[string]any
BenchmarkOutgoing/pack_map[string]any-16                             	  943662	      1342 ns/op
BenchmarkOutgoing/pack_map[string]string
BenchmarkOutgoing/pack_map[string]string-16                          	 3308748	       375.1 ns/op
BenchmarkOutgoing/pack_map[string]any_(any_always_string)
BenchmarkOutgoing/pack_map[string]any_(any_always_string)-16         	  873051	      1281 ns/op
BenchmarkOutgoing/pack_map[string][]string
BenchmarkOutgoing/pack_map[string][]string-16                        	  803454	      1547 ns/op
BenchmarkOutgoing/pack_map[string]any_(any_always_[]string)
BenchmarkOutgoing/pack_map[string]any_(any_always_[]string)-16       	  825868	      1558 ns/op
BenchmarkOutgoing/pack_[]byte
BenchmarkOutgoing/pack_[]byte-16                                     	45481484	        34.66 ns/op


### AFTER ###

BenchmarkOutgoing
BenchmarkOutgoing/pack_[]any
BenchmarkOutgoing/pack_[]any-16                                      	 5936780	       178.7 ns/op
BenchmarkOutgoing/pack_[]bool
BenchmarkOutgoing/pack_[]bool-16                                     	 2564455	       472.9 ns/op
BenchmarkOutgoing/pack_map[string]any
BenchmarkOutgoing/pack_map[string]any-16                             	 3001713	       407.4 ns/op
BenchmarkOutgoing/pack_map[string]string
BenchmarkOutgoing/pack_map[string]string-16                          	 3258835	       386.6 ns/op
BenchmarkOutgoing/pack_map[string]any_(any_always_string)
BenchmarkOutgoing/pack_map[string]any_(any_always_string)-16         	 3530245	       374.7 ns/op
BenchmarkOutgoing/pack_map[string][]string
BenchmarkOutgoing/pack_map[string][]string-16                        	  857400	      1437 ns/op
BenchmarkOutgoing/pack_map[string]any_(any_always_[]string)
BenchmarkOutgoing/pack_map[string]any_(any_always_[]string)-16       	 1801963	       584.2 ns/op
BenchmarkOutgoing/pack_[]byte
BenchmarkOutgoing/pack_[]byte-16                                     	35928717	        39.06 ns/op
```

robsdedude (approved):
:horse_racing::dash: 

Thanks again for taking the time to craft this PR :bow:

zolstein:
Unfortunately, my experience is that it's _quite_ hard to optimize iterating a map with reflection - it's always much faster if you can statically convert to a known map type.

I did find a couple other optimizations, though I'm not sure if you'd want to take them.

The first is that I noticed that using MapIter.Value only allocates when the map values are not pointer-types. Therefore, you can branch on the map value type and avoid allocating a slot for the value if it has pointers, otherwise use the code as written.

```golang
	if isPointerType(t.Elem()) {
		r := v.MapRange()
		for r.Next() {
			key.SetIterKey(r)
			o.packer.String(key.String())
			o.packX(r.Value().Interface())
		}
	} else {
		value := reflect.New(t.Elem()).Elem()
		r := v.MapRange()
		for r.Next() {
			key.SetIterKey(r)
			value.SetIterValue(r)
			o.packer.String(key.String())
			o.packX(value.Interface())
		}
	}
```

The second is that you can put an extra string field in the `outgoing` struct, and use it as the place to store the key value. This lets you avoid an allocation per-map, which is especially useful if there are nested structures. However, the naive version of this breaks if the map has keys that are a new-type of string, and the only workaround I found for this required a bit of unsafe code.

```golang
	key := reflect.NewAt(t.Key(), unsafe.Pointer(&o.stringSlot)).Elem()
```

zolstein:
Also, FYI, it's often helpful to include the `-benchmem` flag when benchmarking, so you also get data on allocations. That's especially true here, since allocations drive most of the performance difference.

robsdedude:
Great pointers :clap: :rocket: Thank you very much once again :bow: 

I actually managed to squeeze a bit more performance out of the packing code by avoiding some unnecessary jumping back and forth between `reflect.Value` and `any`.

neo4j/internal/bolt/outgoing.go
```diff
@@ -490,6 +490,27 @@ func (o *outgoing) packV(v reflect.Value) {
 		}
 	case reflect.Struct:
 		o.packStruct(v.Interface())
+	case reflect.Slice:
+		switch v.Type().Elem().Kind() {
+		case reflect.Uint8:
+			o.packer.Bytes(v.Bytes())
+			return
+		case reflect.Int, reflect.Int64, reflect.String, reflect.Float64, reflect.Interface:
```
zolstein: I think the way you're checking this case is wrong - you want to check if the type is _exactly_ `[]int` or `[]int64` or ..., not just if it's a slice with the right kind. If, say, someone declares `type ID int` and passes in a `[]ID` then it will trigger this case, but it won't get to go through the fast-path in packX. You can check type equality by pre-computing the types you care about. I think `reflect.TypeOf((*[]int)(nil)).Elem()` is the best way, until you can use `reflect.TypeFor` in `1.22`.

neo4j/internal/bolt/outgoing.go
```diff
@@ -490,6 +490,27 @@ func (o *outgoing) packV(v reflect.Value) {
...
+			if v.Len() > 5 {
```
zolstein: This probably bears actually measuring, at least on one machine. My intuition is that the crossover point at which it's worth taking the allocation is likely to be either "v.Len() > 0" or "basically never."

If I remember correctly what reflection is doing under the hood, there's actually not _too_ much overhead reflectively iterating over and indexing into a slice, since it can just track the values as pointers into the slice until you convert to an interface. So the fast-paths mostly just save you looking up the type of the value. (But the branch predictor might make that easy anyway?) Maybe the better solution here is to have fast-paths here that loop through the slices reflectively, but handle the types directly? 

I had looked a bit ago at replacing the `packX` logic with something that worked predominately on `reflect.Value` objects and avoided converting back to interfaces. If you're looking at having a `packV` anyway, it might be better to make `packX` a really thin wrapper and try to move most of the logic and fast-pathing directly here, rather than duplicating most of the logic and moving back and forth between them.

zolstein: Actually, I took a crack at profiling this, and, at least on my machine:

1. Going through the fast-path in packX is, in fact, notably faster.
2. It works out that "v.Len() > 0" is the closer bet than "never' or even "v.Len() > 5".
3. The Interface() call didn't allocate, which I honestly can't quite explain.
4. The version of the optimization I suggested didn't seem to help.

So... nice find. Personally, I'd stick with it and probably change it to "v.Len() > 0" unless there's some other pathological case. In my testing, testing the actual crossover point was move like "v.Len() > 1" but at that point the difference was ~1ns.

robsdedude: "arbitrary guess" is a bit of a lie :sweat_smile: I actually did some benchmarking to determine a rough value that seemed to be close to the cross-over point where the allocation gets amortized (at least on my machine).The 

>Interface() call didn't allocate, which I honestly can't quite explain.

Now this is most peculiar :open_mouth: because in my testing, that call *did* allocate. But I assume the fact that it didn't for you explains why the cross-over point is `> 0` for you, but was higher for me. For a slice with only 1 element, taking the `v.Interface()` route makes the packing more than 2 as slow on my system. So I'd rather not :upside_down_face:.

For the record: I tested on Go 1.23.0 on linux amd64

robsdedude: For that reason, I decided to stick with the minimum size check being some small number as apparently on some systems (like mine) not having it makes things considerably worse while it's not making much of a difference on other systems (like yours).

robsdedude:
Here are some benchmarks of the current state of the PR
```
# +++++ BEFORE +++++
BenchmarkOutgoing
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________-16         	  632931	      1592 ns/op	     541 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_-16         	  795153	      1551 ns/op	     539 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[]bool__________________________________________
BenchmarkOutgoing/pack_[]bool__________________________________________-16         	 2363528	       467.3 ns/op	      98 B/op	      13 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]string____________________
BenchmarkOutgoing/pack_map[string]map[string]string____________________-16         	 1648632	       738.9 ns/op	     378 B/op	       3 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)-16         	  648488	      1777 ns/op	     742 B/op	      17 allocs/op
BenchmarkOutgoing/pack_[]*int__________________________________________
BenchmarkOutgoing/pack_[]*int__________________________________________-16         	 4342869	       272.4 ns/op	     145 B/op	       4 allocs/op
PASS


# +++++ AFTER +++++
BenchmarkOutgoing
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________-16         	  896643	      1296 ns/op	     494 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_-16         	  896656	      1484 ns/op	     350 B/op	       0 allocs/op
BenchmarkOutgoing/pack_[]bool__________________________________________
BenchmarkOutgoing/pack_[]bool__________________________________________-16         	 8330124	       133.1 ns/op	      73 B/op	       0 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]string____________________
BenchmarkOutgoing/pack_map[string]map[string]string____________________-16         	 1588008	       748.1 ns/op	     333 B/op	       2 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)-16         	 1493149	       804.4 ns/op	     353 B/op	       2 allocs/op
BenchmarkOutgoing/pack_[]*int__________________________________________
BenchmarkOutgoing/pack_[]*int__________________________________________-16         	 8745207	       136.0 ns/op	     109 B/op	       0 allocs/op
PASS
```

<details>

<summary>neo4j/internal/bolt/outgoing_bench_test.go</summary>

```go
package bolt

import (
	"github.com/neo4j/neo4j-go-driver/v5/neo4j/internal/packstream"
	"testing"
)

func BenchmarkOutgoing(outer *testing.B) {
	type workload struct {
		description string
		data        any
	}

	type NewAny any

	someInt := 123456789

	workloads := []workload{
		{
			"pack [][]any (any always int)                        ",
			[][]any{
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
			},
		},

		{
			"pack [][]NewAny (NewAny new-type of any, always int) ",
			[][]NewAny{
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
			},
		},

		{
			"pack []bool                                          ",
			[]bool{true, false, true, false, true, true, true, true, false, false, true, false, false},
		},

		{
			"pack map[string]map[string]string                    ",
			map[string]map[string]string{
				"hello": {"world": "world", "oh": "oh", "hello": "world"},
				"int":   {"1": "1", "2": "2", "3": "3"},
			},
		},

		{
			"pack map[string]map[string]any (any is always string)",
			map[string]map[string]any{
				"hello": {"world": "world", "oh": "oh", "hello": "world"},
				"int":   {"1": "1", "2": "2", "3": "3"},
			},
		},

		{
			"pack []*int                                          ",
			[]*int{&someInt, &someInt, &someInt, &someInt},
		},
	}

	for _, load := range workloads {
		outer.Run(load.description, func(inner *testing.B) {
			out := &outgoing{
				chunker:   newChunker(),
				packer:    packstream.Packer{},
				onPackErr: func(err error) { inner.Error(err) },
			}
			for _i := 0; _i < inner.N; _i++ {
				out.packX(load.data)
			}
		})
	}
}
```

</details>

zolstein:
@robsdedude A note on your benchmark code - you're not resetting the buffer inside the Packer between benchmark iterations, which means the work done between iterations isn't consistent - some iterations will need to resize the buffer, others won't. Bytes allocated theoretically average out, though number of allocations won't. IMO it's better to pre-allocate a buffer big enough to avoid resizes and reset it between runs. That way, you can isolate the performance of the serialization code. You also might want to trigger a GC (and reset the timer) between benchmarks.

```golang
buffer := make([]byte, 0, 1024*1024)
...
	out := &outgoing{
		chunker:   newChunker(),
		packer:    packstream.Packer{},
		onPackErr: func(err error) { inner.Error(err) },
	}
	runtime.GC()
	inner.ResetTimer()
	for _i := 0; _i < inner.N; _i++ {
		out.packer.Begin(buffer)
		out.packX(load.data)
	}
```

robsdedude:
Valid points, thanks for the feedback. While it skewed the absolute numbers, I believe that the relative comparison is still valid. After all, the amortized cost of required reallocations of the byte buffer is in `O(1)`. In my test runs, the benchmark indicated the same number or allocs/op regardless if with or without your changes. I might have gotten lucky here, or the fact that the times between reallocations grows exponentially makes it negligible.

Again, I much appreciate your input. I've learned a lot and had a good time digging deep into the Go world :bow: 

Either way, here are the corrected results

## Flawed Benchmark Results (as before, for comparison)
<details>

### Before
```
BenchmarkOutgoing
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________-16         	  784249	      1528 ns/op	     545 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_-16         	  852247	      1501 ns/op	     513 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[]bool__________________________________________
BenchmarkOutgoing/pack_[]bool__________________________________________-16         	 2698767	       441.5 ns/op	      87 B/op	      13 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]string____________________
BenchmarkOutgoing/pack_map[string]map[string]string____________________-16         	 1623920	       717.8 ns/op	     382 B/op	       3 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)-16         	  698413	      1692 ns/op	     719 B/op	      17 allocs/op
BenchmarkOutgoing/pack_[]*int__________________________________________
BenchmarkOutgoing/pack_[]*int__________________________________________-16         	 4519238	       256.0 ns/op	     140 B/op	       4 allocs/op
```

### After
```
BenchmarkOutgoing
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________-16         	  891396	      1226 ns/op	     496 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_-16         	  905594	      1501 ns/op	     347 B/op	       0 allocs/op
BenchmarkOutgoing/pack_[]bool__________________________________________
BenchmarkOutgoing/pack_[]bool__________________________________________-16         	 8906940	       124.8 ns/op	      86 B/op	       0 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]string____________________
BenchmarkOutgoing/pack_map[string]map[string]string____________________-16         	 1473708	       739.3 ns/op	     357 B/op	       2 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)-16         	 1545484	       804.4 ns/op	     342 B/op	       2 allocs/op
BenchmarkOutgoing/pack_[]*int__________________________________________
BenchmarkOutgoing/pack_[]*int__________________________________________-16         	 8766512	       129.3 ns/op	     109 B/op	       0 allocs/op
```

</details>


## Fixed Benchmark Results
### Before
```
BenchmarkOutgoing
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________-16         	  755818	      1400 ns/op	     144 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_-16         	  880044	      1408 ns/op	     144 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[]bool__________________________________________
BenchmarkOutgoing/pack_[]bool__________________________________________-16         	 2763163	       419.3 ns/op	      13 B/op	      13 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]string____________________
BenchmarkOutgoing/pack_map[string]map[string]string____________________-16         	 1966670	       605.6 ns/op	      80 B/op	       3 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)-16         	  725830	      1620 ns/op	     432 B/op	      17 allocs/op
BenchmarkOutgoing/pack_[]*int__________________________________________
BenchmarkOutgoing/pack_[]*int__________________________________________-16         	 4844119	       221.7 ns/op	      32 B/op	       4 allocs/op
```

### After
```
BenchmarkOutgoing
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________
BenchmarkOutgoing/pack_[][]any_(any_always_int)________________________-16         	  998196	      1170 ns/op	     144 B/op	       6 allocs/op
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_
BenchmarkOutgoing/pack_[][]NewAny_(NewAny_new-type_of_any,_always_int)_-16         	  931497	      1330 ns/op	       0 B/op	       0 allocs/op
BenchmarkOutgoing/pack_[]bool__________________________________________
BenchmarkOutgoing/pack_[]bool__________________________________________-16         	11988564	       132.5 ns/op	      31 B/op	       0 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]string____________________
BenchmarkOutgoing/pack_map[string]map[string]string____________________-16         	 1635100	       653.8 ns/op	      24 B/op	       2 allocs/op
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)
BenchmarkOutgoing/pack_map[string]map[string]any_(any_is_always_string)-16         	 1775896	       670.2 ns/op	      24 B/op	       2 allocs/op
BenchmarkOutgoing/pack_[]*int__________________________________________
BenchmarkOutgoing/pack_[]*int__________________________________________-16         	12874026	       124.9 ns/op	      75 B/op	       0 allocs/op
```

<details>

<summary>fixed neo4j/internal/bolt/outgoing_bench_test.go</summary>

```go
	// ...
	for _, load := range workloads {
		outer.Run(load.description, func(inner *testing.B) {
			out := &outgoing{
				chunker:   newChunker(),
				packer:    packstream.Packer{},
				onPackErr: func(err error) { inner.Error(err) },
			}
			out.packX(load.data)
			data, err := out.packer.End()
			if err != nil {
				inner.Error(err)
			}
			// create buffer large enough to avoid reallocations
			buffer := make([]byte, 0, len(data)*inner.N)
			out.packer.Begin(buffer)
			runtime.GC()
			inner.ResetTimer()
			for _i := 0; _i < inner.N; _i++ {
				out.packX(load.data)
			}
		})
	}
	// ...
```

</details>

StephenCathcart (approved):
Code looks good, nice test coverage 👍

StephenCathcart:
Changes and tests look great, thanks for your help on this @zolstein 🙇

PR #618, opened 2025-01-29
ADR-026 Home Database Cache

StephenCathcart:
Added support for Bolt 5.8, which introduces a couple of useful features. The server now echoes back the resolved home/default database when starting a transaction, along with a connection hint indicating whether server-side routing (SSR) is enabled. If SSR is on, the echoed database can be used with an optimistic home database cache to predict a client-side route (optimistic routing), falling back to SSR if the guess is wrong. This can eliminate a round-trip, but only if all the following conditions are met:

- Server-side routing (SSR) is enabled
- Server supports Bolt 5.8
- A brand-new session (not yet used) starts its first transaction (auto-commit or explicit) without an explicitly set target database


neo4j/internal/pool/pool.go
```diff
@@ -507,3 +514,21 @@ func (p *Pool) deactivateWriter(serverName string, db string) {
 	p.log.Debugf(log.Pool, p.logId, "Deactivating writer %s for database %s", serverName, db)
 	p.router.InvalidateWriter(db, serverName)
 }
+
+func (p *Pool) updateCacheState() {
```
robsdedude, line 529: Functionally correct as far as I can tell. But this has optimization potential. The `server` could instead track the number of connections with and without SSR using two atomics. This is faster beacuse 
 *  not holding a lock on the whole server while iterating over all connections
 * not performing the update operation after every pool borrow/return, but only when a connection is being opened/closed

neo4j/driver_with_context.go
```diff
@@ -295,12 +302,12 @@ type sessionRouter interface {
 	// note: bookmarks are lazily supplied, only when a new routing table needs to be fetched
 	// this is needed because custom bookmark managers may provide bookmarks from external systems
 	// they should not be called when it is not needed (e.g. when a routing table is cached)
-	GetOrUpdateReaders(ctx context.Context, bookmarks func(context.Context) ([]string, error), database string, auth *idb.ReAuthToken, boltLogger log.BoltLogger) ([]string, error)
+	GetOrUpdateReaders(ctx context.Context, bookmarks func(context.Context) ([]string, error), database string, auth *idb.ReAuthToken, boltLogger log.BoltLogger, useHomeDbGuess bool) ([]string, error)
```
robsdedude: Potential for refactoring: the `useHomeDbGuess` flag is probably better called `isHomeDbGuess` or just `guessedDb` or such. Further, I think it would go great together with the `database string` into a single struct as this data is always passed around together until eventually it's broken apart and only the string (or  `""`) is passed further down the stack.

neo4j/internal/homedb/cache.go
```diff
@@ -0,0 +1,243 @@
...
+	"sort"
+	"sync"
+	"time"
+)
+
+// DefaultCacheMaxSize defines the maximum number of entries the cache can hold.
+const DefaultCacheMaxSize = 1000
+
+// cachePruneFraction defines the fraction of entries to prune when the cache exceeds its size.
+const cachePruneFraction = 0.1
```
robsdedude: [The ADR](https://github.com/neo-technology/drivers-adr/blob/main/adrs/026-home-db-resolution.md#cache-replacement-policy) states that (and how) `cachePruneFraction` should depend on the size of the cache. Currently, the cache size is fixed, but the cache is written in a way that it can easily be changed from the outside (`NewCache` takes a `maxSize`).

In order for this relation between the numbers not to get lost, I recommend to do one of the following:
 * compute the prune factor dynamically in `NewCache`
 * make `NewCache` not take a `maxSize`, but instead always use `DefaultCacheMaxSize`. In this case, you can pre-compute `cachePruneFraction` and use a constant like it's right now. When using this option, I suggest to leave a comment at the const definition that explains how to update the value should `DefaultCacheMaxSize` be changed (i.e., put the formula there).

neo4j/internal/homedb/cache.go
```diff
@@ -0,0 +1,243 @@
...
+	entry.lastUsed = time.Now()
+	return entry.database, true
+}
+
+// Set adds or updates an entry in the cache.
+// If the cache exceeds its max size, it prunes the least recently used entries.
+func (c *Cache) Set(user string, database string) {
+	if !c.IsEnabled() {
+		return
+	}
```
robsdedude: If I read the code right, when SSR is disabled/unknown, the cache is not updated even if we learn about about a user -> homeDD mapping.  We decided to track the homeDBs regardless of SSR availability (s. [ADR](https://github.com/neo-technology/drivers-adr/blob/main/adrs/026-home-db-resolution.md#maintaining-the-cache-when-ssr-is-disabled)).

I thought there's a TestKit test for that :eyes: If not, could you add one?

neo4j/internal/homedb/cache.go
```diff
@@ -0,0 +1,243 @@
...
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	delete(c.cache, user)
+}
+
+// InvalidateDatabase removes all entries that reference a specific database.
+func (c *Cache) InvalidateDatabase(database string) {
```
robsdedude: Nit: dead code (unused method)

neo4j/session_with_context.go
```diff
@@ -854,3 +941,14 @@ func validateTransactionConfig(config TransactionConfig) error {
...
+func computeCacheKey(ctx context.Context, token *idb.ReAuthToken, cache *homedb.Cache, impersonatedUser string) (string, error) {
+	if token.FromSession {
+		t, err := token.Manager.GetAuthToken(ctx)
+		if err != nil {
+			return "", err
+		}
+		return cache.ComputeKey(impersonatedUser, t, token.FromSession), nil
+	}
+	return cache.ComputeKey(impersonatedUser, AuthToken{}, token.FromSession), nil
+}
```
robsdedude: Refactoring potential: `ComputeKey` could take a `sessionAuth *auth.Token` instead of `auth auth.Token, fromSession bool`. Because currently, when `fromSession == true`, the token is not used at all and in fact at the call site (here) it's being filled with a placeholder `AuthToken{}`. At that point, just sending in `nil` seems like the cheaper and cleaner design to indicate "yeah, no, I don't have a session auth token for you".

neo4j/internal/homedb/cache.go
```diff
@@ -0,0 +1,243 @@
...
+
+	for user, entry := range c.cache {
+		if entry.database == database {
+			delete(c.cache, user)
+		}
+	}
+}
+
+// ComputeKey generates a cache key based on user impersonation, auth token, and session context.
+func (c *Cache) ComputeKey(impersonatedUser string, auth auth.Token, fromSession bool) string {
```
robsdedude: Please make sure there are no cache key collisions.
Here's one example that fails but shouldn't:

```go
	outer.Run("collision", func(t *testing.T) {
		cache := &Cache{}
		token1 := auth.Token{
			Tokens: map[string]any{
				"scheme": "custom",
				"parameters": map[string]any{
					"key1:fun": "value1",
				},
			},
		}
		token2 := auth.Token{
			Tokens: map[string]any{
				"scheme": "custom",
				"parameters": map[string]any{
					"key1": "fun:value1",
				},
			},
		}
		key1 := cache.ComputeKey("", token1, true)
		key2 := cache.ComputeKey("", token2, true)
		testutil.AssertNotDeepEquals(t, key1, key2)
	})
```

neo4j/internal/homedb/cache.go
```diff
@@ -0,0 +1,243 @@
...
+			}
+			return fmt.Sprintf("%s:%s%s:%v%s%s%s:%s",
+				keyScheme, scheme,
+				keyPrincipal, auth.Tokens[keyPrincipal],
+				credentialString, realmString,
+				keyParameters, orderedParams)
+		}
+	}
+	// If no specific key could be determined, return a default
+	return "DEFAULT"
```
robsdedude: I think the fall-through (i.e., I can't make sense of this token, hence I can't optimize the key) should be to stupidly serialize the whole token into a key (similar to or even the same way as for the custom scheme case). We can't conclude that the token we don't understand means "driver level user".

neo4j/internal/homedb/cache.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// NewCache creates and returns a new cache instance with the given max size.
+func NewCache(maxSize int) (*Cache, error) {
+	if maxSize <= 0 {
+		return nil, &errorutil.UsageError{Message: "Maximum cache size must be greater than 0"}
+	}
+	return &Cache{
+		maxSize: maxSize,
+		cache:   make(map[string]*cacheEntry),
```
robsdedude:
```suggestion
		cache:   make(map[string]*cacheEntry, maxSize + 1),
```
Not sure if worth it... Saves some potential allocations later down the road at the cost of a higher memory footprint (but probably not by much).

neo4j/internal/router/router.go
```diff
@@ -176,8 +176,13 @@ func (r *Router) getOrUpdateTable(ctx context.Context, bookmarksFn func(context.
...
 		unlock.Do(r.dbRoutersMut.Unlock)
-
-		table, err := r.updateTable(ctx, bookmarksFn, database, auth, boltLogger, dbRouter)
+		// Use an empty string for updating the routing table if the home database is a guess,
+		// as we cannot guarantee the guess is correct.
+		targetDatabase := database
+		if useHomeDbGuess {
+			targetDatabase = ""
+		}
+		table, err := r.updateTable(ctx, bookmarksFn, targetDatabase, auth, boltLogger, dbRouter)
```
robsdedude: Nit: maybe add a test that covers this behavior

neo4j/session_with_context.go
```diff
@@ -234,21 +240,30 @@ func newSessionWithContext(
...
+		homeDbGuess:   homeDbGuess,
 		sleep:         racing.Sleep,
 		log:           logger,
 		logId:         logId,
 		throttleTime:  time.Second * 1,
 		fetchSize:     fetchSize,
 		auth:          token,
 	}
+	session.pinHomeDatabaseCallback = func(database string) {
+		session.pinHomeDatabase(ctx, database)
```
robsdedude, line 274: This is binding the context meant for session creating to an event that could be happening much later. The pinning of the home db (among other events) can happen when the `BEGIN`'s/`RUN`'s `SUCCESS` message is received. Those should be governed by the context passed to `sessioin.Run`, `session.BeginTransaction` or similar.

neo4j/session_with_context.go
```diff
@@ -545,49 +564,105 @@ func (s *sessionWithContext) getServers(mode idb.AccessMode) func() []string {
...
+	conn.SetPinHomeDatabaseCallback(s.pinHomeDatabaseCallback)
+
 	// Select database on server
 	if s.config.DatabaseName != idb.DefaultDatabase {
-		dbSelector, ok := conn.(idb.DatabaseSelector)
-		if !ok {
-			s.pool.Return(ctx, conn)
-			return nil, &UsageError{Message: "Database does not support multi-database"}
+		if err := s.selectDatabase(ctx, conn); err != nil {
+			return nil, err
```
robsdedude, line 629: Before baling out, we need to return the connection we've already borrowed. Else, we'll clog up the pool.

**EDIT**: I just saw that `selectDatabase` actually returns the connection on error. So no bug here. But hard to follow code, I'd argue. Smells like spooky action at a distance to me.

neo4j/session_with_context.go
```diff
@@ -545,49 +564,105 @@ func (s *sessionWithContext) getServers(mode idb.AccessMode) func() []string {
...
+	if s.config.DatabaseName != "" {
+		if err := s.resolveHomeDatabase(ctx); err != nil {
+			return nil, false, err
+		}
+		serverList, err := s.getOrUpdateServers(ctx, mode, false)
+		return serverList, false, err
+	}
+
+	// Use cached home database if available
+	if s.homeDbGuess != "" && s.router.GetTable(s.homeDbGuess) != nil && s.cache.IsEnabled() {
```
robsdedude:
```suggestion
	if s.homeDbGuess != "" && s.cache.IsEnabled() {
```
`getOrUpdateServers` is going to call `s.router.GetOrUpdateReaders/Writers` regardless which will fetch a RT if none exists. So why bother with this check that could be outdated by the time the router gets to decided whether it needs to fetch a RT or not. Further, the existence of a RT doesn't guarantee that it's valid. The TTL could've expired, there could be no readers/writers left, etc.

neo4j/session_with_context.go
```diff
@@ -545,49 +564,105 @@ func (s *sessionWithContext) getServers(mode idb.AccessMode) func() []string {
...
+		serverList, err := s.getOrUpdateServers(ctx, mode, true)
+		return serverList, true, err
+	}
+
+	// Resolve home database if cache is unavailable, there's no guess, or a table doesn't exist.
+	if err := s.resolveHomeDatabase(ctx); err != nil {
+		return nil, false, err
+	}
+	serverList, err := s.getOrUpdateServers(ctx, mode, s.cache.IsEnabled())
```
robsdedude:
```suggestion
	serverList, err := s.getOrUpdateServers(ctx, mode, false)
```
At this point, we've explicitly resolved and pinned the homeDB. We don't care about the cache any longer. It also didn't align with the return value below claiming that no cache was used.

neo4j/session_with_context.go
```diff
@@ -545,49 +564,105 @@ func (s *sessionWithContext) getServers(mode idb.AccessMode) func() []string {
...
+	return serverList, false, err
+}
+
+// borrowConnection requests a connection from the pool using the provided `serverList`.
+func (s *sessionWithContext) borrowConnection(ctx context.Context, serverList []string, livenessCheckTimeout time.Duration) (idb.Connection, error) {
+	conn, err := s.pool.Borrow(ctx, func() []string { return serverList }, s.driverConfig.ConnectionAcquisitionTimeout != 0, s.config.BoltLogger, livenessCheckTimeout, s.auth)
+	return conn, err
+}
```
robsdedude:
```suggestion
func (s *sessionWithContext) borrowConnection(
	ctx context.Context,
	serverList []string,
	livenessCheckTimeout time.Duration,
) (idb.Connection, error) {
	return s.pool.Borrow(
		ctx,
		func() []string { return serverList },
		s.driverConfig.ConnectionAcquisitionTimeout != 0,
		s.config.BoltLogger,
		livenessCheckTimeout,
		s.auth,
	)
}
```
:broom:

neo4j/session_with_context.go
```diff
@@ -545,49 +564,105 @@ func (s *sessionWithContext) getServers(mode idb.AccessMode) func() []string {
...
+
+// selectDatabase ensures the correct database is selected on the connection.
+func (s *sessionWithContext) selectDatabase(ctx context.Context, conn idb.Connection) error {
+	dbSelector, ok := conn.(idb.DatabaseSelector)
+	if !ok {
+		s.pool.Return(ctx, conn)
+		return &UsageError{Message: "Database does not support multi-database"}
```
robsdedude: I know this code has been here before. I think this go into a separate PR, if we decided we want to do it. While I see that code, I remember we have a specific error for such things that gets turned into a `UsageError` by `errorutil.WrapError` anyway:
```suggestion
		err := &db.FeatureNotSupportedError{
			Server: conn.ServerName(), Feature: "multi-database", Reason: "requires at least server v4",
		}
		s.pool.Return(ctx, conn)
		return err
```

neo4j/session_with_context.go
```diff
@@ -725,7 +800,7 @@ func (s *sessionWithContext) getServerInfo(ctx context.Context) (ServerInfo, err
 	if err := s.resolveHomeDatabase(ctx); err != nil {
 		return nil, errorutil.WrapError(err)
 	}
-	_, err := s.getOrUpdateServers(ctx, idb.ReadMode)
+	_, err := s.getOrUpdateServers(ctx, idb.ReadMode, false)
```
robsdedude: `getServerInfo`, is no exception from this new feature. It, too, can and should make use of the home db cache.

Worth writing a TestKit test for, I think :memo: 

N.B. technically, `verifyAuthentication` should also make use of the cache. But the way it's exposed to the user this function will only be called with the session configured with `"system"` as DB. But that again is making assumptions cross concern boundaries, which is a recipe for regressions (spooky action at a distance).

neo4j/session_with_context.go
```diff
@@ -784,12 +859,24 @@ func (s *sessionWithContext) resolveHomeDatabase(ctx context.Context) error {
...
+	s.pinHomeDatabase(ctx, defaultDb)
 	return nil
 }
 
+func (s *sessionWithContext) pinHomeDatabase(ctx context.Context, database string) {
+	if !s.resolveHomeDb {
+		return
+	}
+
+	key, _ := computeCacheKey(ctx, s.auth, s.cache, s.config.ImpersonatedUser)
```
robsdedude: :eyes: I spy with my little eye a swallowed error. Is this on purpose? Do we want to continue with a key that might be `""`?

testkit-backend/backend.go
```diff
@@ -1379,15 +1382,33 @@ func getAuth(authTokenMap map[string]any) (neo4j.AuthToken, error) {
...
+			parameters = v
+			if err := patchNumbersInMap(parameters); err != nil {
+				return neo4j.AuthToken{}, err
+			}
+		}
+
+		var scheme, principal, credentials, realm string
+		if v, ok := authTokenMap["scheme"].(string); ok {
+			scheme = v
+		}
```
robsdedude: Refactoring potential: 4 times basically the same 3 lines. Something along the lines of

```go
func mapGetString(data map[string]any, key string) string {
	out, _ := data[key].(string)
	return out
}
```

neo4j/session_with_context.go
```diff
@@ -545,49 +564,105 @@ func (s *sessionWithContext) getServers(mode idb.AccessMode) func() []string {
...
+		s.log.Debugf(log.Session, s.logId, "connection acquisition user-provided deadline is: %s", deadline)
+	}
+	return ctx, nil
+}
+
+// getServerList resolves the server list based on the session configuration.
+// It returns a list of servers, a boolean indicating whether the home database guess was used, and an error if resolution fails.
+func (s *sessionWithContext) getServerList(ctx context.Context, mode idb.AccessMode) ([]string, bool, error) {
+	if s.config.DatabaseName != "" {
+		if err := s.resolveHomeDatabase(ctx); err != nil {
```
robsdedude: Why are we calling `resolveHomeDatabase` when there is an explicit database configured on the session?

neo4j/session_with_context.go
```diff
@@ -784,12 +859,24 @@ func (s *sessionWithContext) resolveHomeDatabase(ctx context.Context) error {
...
 
+func (s *sessionWithContext) pinHomeDatabase(ctx context.Context, database string) {
+	if !s.resolveHomeDb {
+		return
+	}
+
+	key, _ := computeCacheKey(ctx, s.auth, s.cache, s.config.ImpersonatedUser)
+	s.cache.Set(key, database)
+
+	s.log.Debugf(log.Session, s.logId, "Resolved home database, uses db '%s'", database)
```
robsdedude, line 918: You know I like myself some good logging :smile: Optional suggestion: a bit more logging in the session. E.g., when we decide to explicitly resolve the home database, ... It's up to you and ofc. should't become too noisy.

robsdedude (changes requested):

neo4j/session_with_context.go
StephenCathcart: hangover from original code - `resolveHomeDatabase` would just return early here but it's redundant. Removed.

neo4j/driver_with_context.go
StephenCathcart: Agreed, I've lumped these into a database selection struct in `idb`.

neo4j/internal/homedb/cache.go
StephenCathcart: Yep you're correct this check shouldn't be there. Happy to add a test case for that.

neo4j/internal/homedb/cache.go
StephenCathcart: Refactored to compute prune factor dynamically in `NewCache`

neo4j/session_with_context.go
StephenCathcart: Agree - updated

neo4j/internal/homedb/cache.go
StephenCathcart: refactored and added a `serializeMap`, which is also now used for custom params 👍

neo4j/internal/homedb/cache.go
StephenCathcart: fixed - the key of the maps are now wrapped in brackets `<key>:value`

neo4j/internal/router/router.go
StephenCathcart: added

neo4j/internal/homedb/cache.go
StephenCathcart: probs worth the footprint, updated 👍

neo4j/session_with_context.go
StephenCathcart: good suggestion

neo4j/session_with_context.go
StephenCathcart: Thank you I was debating adding the feature to `getServerInfo` and `verifyAuthentication` but wasn't sure 👍

neo4j/session_with_context.go
StephenCathcart: Yeah looks like this happens if we fail to get the auth info from the manager, and at that point, we probs don't want to use the empty key for the cache. Added a check and logging for when it errors out.

neo4j/session_with_context.go
StephenCathcart: good catch

neo4j/session_with_context.go
StephenCathcart: added additional logging 👍

neo4j/session_with_context.go
StephenCathcart: Might need to discuss this code a bit further, without the check TestKit fails along with a unit test for checking this scenario. As to why the code was there, it'll be from following this part of the ADR: "If a routing table for the cached db is not available, or there is no cached db"

neo4j/session_with_context.go
StephenCathcart: Yeah, this is old `getConnection` code, it took me a second to spot that `selectDatabase` does a return too. I've refactored the returning logic for the caller to handle.

neo4j/session_with_context.go
StephenCathcart: Added onRoutingTableUpdated callback to router for handling this logic rather than relying on GetTable 👍 .

neo4j/internal/pool/pool.go
StephenCathcart: As discussed, exploring pros/cons of the ssr atomic approach 👍

neo4j/session_with_context.go
```diff
@@ -545,49 +583,126 @@ func (s *sessionWithContext) getServers(mode idb.AccessMode) func() []string {
...
+	if deadline, ok := ctx.Deadline(); ok {
+		s.log.Debugf(log.Session, s.logId, "connection acquisition user-provided deadline is: %s", deadline)
+	}
+	return ctx, nil
+}
+
+// getServerList resolves the server list based on the session configuration.
+// It returns a list of servers, a boolean indicating whether the home database guess was used, and an error if resolution fails.
+func (s *sessionWithContext) getServerList(ctx context.Context, mode idb.AccessMode) ([]string, bool, error) {
+	if s.config.DatabaseName != "" {
```
robsdedude:
```suggestion
	if !s.resolveHomeDb {
```
Should be equivalent, but 
 * faster when either connected to old servers that do not resolve the home db (pre 4.3, I believe) or when using a direct driver (`bolt+[s[sc]]://...`) where the home db resolution is a no-op.
 * avoids logging `"No valid cached home database guess available, resolving home database"` while the resolution will be a no-op.

StephenCathcart: yeah nice change, can confirm tests still pass too

neo4j/internal/pool/pool.go
robsdedude: https://github.com/StephenCathcart/neo4j-go-driver/pull/2

PR #619, opened 2025-02-06
ADR 30: Bolt Handshake Manifest v1

StephenCathcart:
Implemented the new Bolt Protocol Handshake Manifest v1 negotiation.

The client now sends a handshake proposal with a manifest marker `(00 00 01 FF)` in addition to the legacy protocol proposals.

When the server selects the manifest negotiation, the client reads a variable-length count of supported protocol versions, the list of offered protocol versions, and a capability bit mask (all encoded as Base128 varints). The client then selects an acceptable protocol version along with capabilities and sent immediately before the HELLO message.

```
2025-02-06 09:47:07.076   BOLT  C: <MAGIC> 0X6060B017
2025-02-06 09:47:07.076   BOLT  C: <HANDSHAKE> 0X000001FF 0X00070705 0X00020404 0X00000003
2025-02-06 09:47:07.079   BOLT  S: <HANDSHAKE> 0X000001FF [2] 0x000707F2 0x00070705 0X00
2025-02-06 09:47:07.079   BOLT  C: <HANDSHAKE> 0X00000705 0X00
2025-02-06 09:47:07.079   BOLT  C: HELLO {"bolt_agent":{"language":"Go/go1.20.3","platform":"darwin; amd64","product":"neo4j-go/5.27.0"},"user_agent":"Go Driver/5.27.0"}
```


neo4j/internal/bolt/connect.go
```diff
@@ -35,12 +37,18 @@ type protocolVersion struct {
 	back  byte // Number of minor versions back
 }
 
-// Supported versions in priority order
+func (p *protocolVersion) formatProtocol() string {
+	return fmt.Sprintf("0x%04X%02X%02X", p.back, p.minor, p.major)
```
robsdedude:
```suggestion
	return fmt.Sprintf("%#04X%02X%02X", p.back, p.minor, p.major)
```
For consistent formatting (`0X0000...` instead of `0x0000...`).

neo4j/internal/bolt/connect.go
robsdedude, line 1: https://github.com/neo4j/neo4j-go-driver/pull/619/files#diff-d9c2e9af24cb7e2f4973aec46b487836c814441d4b5e3254ab1889127785a720R125-R128

This check could be moved up as it's only sensible to be performed on the original response (the first 4 bytes received from the server). But where it is right now, `major` and `minor` might have already been overwritten by `performManifestNegotiation`.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+func readCapabilityMask(ctx context.Context, r racing.RacingReader, serverName string, errorListener ConnectionErrorListener) (uint64, []byte, error) {
+	capMask, err := readVarInt(ctx, r)
+	if err != nil {
+		errorListener.OnDialError(ctx, serverName, err)
+		return 0, nil, fmt.Errorf("failed to read capability mask: %w", err)
+	}
+	capBytes, err := encodeVarInt(capMask)
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to encode capability mask: %w", err)
+	}
```
robsdedude: I think we can save time and an error branch that's (barring bugs) unreachable if we make `readVarInt` also return the raw bytes.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+	if err != nil {
+		invalidHandshake := []byte{0x00, 0x00, 0x00, 0x00, 0x00} // 4 bytes for version + 1 byte for capabilities.
+		if _, err := conn.Write(invalidHandshake); err != nil {
+			errorListener.OnDialError(ctx, serverName, err)
+		}
+		return 0, 0, err
+	}
+
+	// Send the handshake confirmation.
+	if err = sendHandshakeConfirmation(ctx, conn, boltLogger, errorListener, serverName, chosen, capBytes); err != nil {
```
robsdedude:
```suggestion
	if err = sendHandshakeConfirmation(ctx, conn, boltLogger, errorListener, serverName, chosen, []byte{0x00}); err != nil {
```
I don't think we should simply echo back whatever capability bytes the server sent us. That way we'd opt into any future capabilities even without changing the driver code. That's error-prone.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+		fmt.Sprintf("%#X", response),
+		count,
+		strings.Join(supportedProtocols, " "),
+		fmt.Sprintf("%#X", capBytes))
+}
+
+// selectProtocol iterates over our protocol proposals (skipping the manifest marker)
+// and returns the first candidate whose major version matches and whose minor version
+// falls within the range offered by the server.
+func selectProtocol(supported []protocolVersion, errorListener ConnectionErrorListener, serverName string) (protocolVersion, error) {
```
robsdedude:
```suggestion
func selectProtocol(supported []protocolVersion) (protocolVersion, error) {
```
Unused parameters. However, maybe `errorListener` should actually be used instead of being removed.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+	logManifestHandshake(boltLogger, response, count, supported, capBytes)
+
+	// Select an acceptable protocol version.
+	chosen, err := selectProtocol(supported, errorListener, serverName)
+	if err != nil {
+		invalidHandshake := []byte{0x00, 0x00, 0x00, 0x00, 0x00} // 4 bytes for version + 1 byte for capabilities.
+		if _, err := conn.Write(invalidHandshake); err != nil {
+			errorListener.OnDialError(ctx, serverName, err)
+		}
+		return 0, 0, err
```
robsdedude: `selectProtocol` does not call `errorListener.OnDialError` on error. This place only calls it if writing `invalidHandshake` errors out. So if `selectProtocol` failes, but `conn.Write(invalidHandshake)` doesn't for what ever reason, we don't call `OnDialError`. I think `conn.Write(invalidHandshake)` failing can be silently ignored. It's just a best-effort to tell the server "yeah no :point_right::point_left: listen... I don't think this us-thing works, sorry, it's me, not you :broken_heart:". If the server doesn't receive it, just closing the connection and returning the error is fine here. However, `selectProtocol` should be calling `OnDialError` when causing an error.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+// and returns the first candidate whose major version matches and whose minor version
+// falls within the range offered by the server.
+func selectProtocol(supported []protocolVersion, errorListener ConnectionErrorListener, serverName string) (protocolVersion, error) {
+	proposals := versions[1:]
+	for _, candidate := range proposals {
+		for _, offer := range supported {
+			if candidate.major == offer.major &&
+				candidate.minor <= offer.minor &&
+				candidate.minor >= (offer.minor-offer.back) {
+				return candidate, nil
```
robsdedude: This is flawed. If the server were to offer `0x00060605 0x00040404` with our current `versions` (which includes `0x00070705`), we'd not be choosing `5.6`, but `4.4`.

This should've been caught by a TestKit test :cold_sweat: time to expand the test suite.

Please add more tests. At least: `sever max version == driver max version`, `>`, and `<`.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+	}
+
+	// Log the complete server handshake message.
+	logManifestHandshake(boltLogger, response, count, supported, capBytes)
+
+	// Select an acceptable protocol version.
+	chosen, err := selectProtocol(supported, errorListener, serverName)
+	if err != nil {
+		invalidHandshake := []byte{0x00, 0x00, 0x00, 0x00, 0x00} // 4 bytes for version + 1 byte for capabilities.
+		if _, err := conn.Write(invalidHandshake); err != nil {
```
robsdedude: Shouldn't all writes be using a `racing.RacingWriter`?

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+
+// readProtocolOfferings reads the number of protocol offerings and returns the count and
+// a slice of supported protocol versions.
+func readProtocolOfferings(ctx context.Context, r racing.RacingReader, serverName string, errorListener ConnectionErrorListener) (uint64, []protocolVersion, error) {
+	count, err := readVarInt(ctx, r)
+	if err != nil {
+		errorListener.OnDialError(ctx, serverName, err)
+		return 0, nil, fmt.Errorf("failed to read manifest protocol count: %w", err)
+	}
+	supported := make([]protocolVersion, count)
```
robsdedude:
```suggestion
	supported := make([]protocolVersion, 0, count)
```
No need to initialize all protocolVersion structs just to then overwrite them further down.
If accepted, also change `supported[i] = protocolVersion{...}` to use `append` instead.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+			errorListener.OnDialError(ctx, serverName, err)
+			return 0, nil, fmt.Errorf("failed to read manifest protocol version: %w", err)
+		}
+		supported[i] = protocolVersion{
+			back:  versionBytes[1],
+			minor: versionBytes[2],
+			major: versionBytes[3],
+		}
+	}
+	return count, supported, nil
```
robsdedude: Why is the `count` returned? Isn't is already implicitly returned as `len(supported)`?

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+	return capMask, capBytes, nil
+}
+
+// logManifestHandshake logs the complete server handshake message for manifest negotiation.
+// It prints the initial response, count of offerings, each supported protocol, and the capability mask.
+func logManifestHandshake(boltLogger log.BoltLogger, response []byte, count uint64, supported []protocolVersion, capBytes []byte) {
+	if boltLogger == nil {
+		return
+	}
+	var supportedProtocols []string
```
robsdedude: Nit: just debug code, but could still be created with the right capacity.

neo4j/internal/bolt/connect.go
```diff
@@ -115,3 +134,181 @@ func Connect(ctx context.Context,
...
+		b := buf[0]
+		result |= uint64(b&0x7F) << shift
+		// The most significant bit is the continuation flag.
+		if b&0x80 == 0 {
+			break
+		}
+		shift += 7
+		if shift >= 64 {
+			return 0, fmt.Errorf("varint too long")
+		}
```
robsdedude: In all realistic scenarios, this is fine. However, note that 64 isn't divisible by 7. The next smaller divisible number is 63. That means, that the most significant bit can never be 1.

neo4j/internal/bolt/connect.go
```diff
@@ -35,12 +37,18 @@ type protocolVersion struct {
...
+}
+
+// versions lists the supported protocol versions in priority order.
+// The first proposal is a marker indicating that the client wishes to use the
+// new manifest-style negotiation.
 var versions = [4]protocolVersion{
+	{major: 0xFF, minor: 0x01, back: 0x00}, // Bolt manifest marker
 	{major: 5, minor: 7, back: 7},
 	{major: 4, minor: 4, back: 2},
-	{major: 4, minor: 1},
```
robsdedude, line 42: The drop of version 4.1 would allow us to remove some `minor` version branches in `bolt4.go`, if we'd wanted to.

neo4j/internal/bolt/connect_test.go
robsdedude, line 1: Optional, but I'd love to see unit tests for the varint read & write functions.

robsdedude (changes requested):

neo4j/internal/bolt/connect.go
robsdedude: Tests added and failing as anticipated 
* https://github.com/neo4j-drivers/testkit/pull/652

neo4j/internal/bolt/connect.go
StephenCathcart: Yup, left this in after a refactor - removed 👍

neo4j/internal/bolt/connect.go
StephenCathcart: moved this check higher up

neo4j/internal/bolt/connect.go
StephenCathcart: I'll look at the branches - would we ever expand this range again in the future to include 4.1, and therefore still need the logic?

neo4j/internal/bolt/connect.go
StephenCathcart: refactoring this whole thing to use `encoding/binary` instead...

neo4j/internal/bolt/connect.go
StephenCathcart: error handling is no longer needed after refactoring.

neo4j/internal/bolt/connect_test.go
StephenCathcart: added

neo4j/internal/bolt/connect.go
StephenCathcart: refactored to handle this

neo4j/internal/bolt/connect.go
robsdedude: I don't think so. Maybe raise the question in stand-up to make sure the rest of the team has the same opinion.

neo4j/internal/bolt/connect.go
StephenCathcart: Can confirm the new Testkit tests are now passing with changes 🍏

neo4j/internal/bolt/connect.go
StephenCathcart: makes sense, refactored 👍

neo4j/internal/bolt/connect.go
StephenCathcart: a branch a unit test have been removed 👍

neo4j/internal/bolt/connect.go
```diff
@@ -260,55 +261,41 @@ func sendHandshakeConfirmation(ctx context.Context, conn io.ReadWriteCloser, bol
...
-			break
-		}
-		shift += 7
-		if shift >= 64 {
-			return 0, fmt.Errorf("varint too long")
+		// If the continuation bit is not set, we've reached the end of the varint.
+		if buf[i]&0x80 == 0 {
+			value, n := binary.Uvarint(buf[:i+1])
+			if n <= 0 {
+				return 0, fmt.Errorf("failed to decode varint")
```
robsdedude: This cannot possibly happen, or am I wrong in thinking that we already make sure that the buffer is at most 10 bytes (`binary.MaxVarintLen64`) long and the last byte does not have the continuation bit set.

neo4j/internal/bolt/connect.go
```diff
@@ -260,55 +261,41 @@ func sendHandshakeConfirmation(ctx context.Context, conn io.ReadWriteCloser, bol
...
-// decoded unsigned integer, or an error if the value is too long or the read fails.
+// readVarInt returns a Base128-encoded variable-length integer from the reader.
 func readVarInt(ctx context.Context, r racing.RacingReader) (uint64, error) {
-	var result uint64
-	var shift uint
-	var buf [1]byte
-	for {
-		_, err := r.Read(ctx, buf[:])
-		if err != nil {
+	var buf [10]byte
```
robsdedude:
```suggestion
	var buf [binary.MaxVarintLen64]byte
```

neo4j/internal/bolt/connect.go
```diff
@@ -260,55 +261,41 @@ func sendHandshakeConfirmation(ctx context.Context, conn io.ReadWriteCloser, bol
...
-			buf = append(buf, b|0x80)
-		} else {
-			buf = append(buf, b)
-			break
-		}
-	}
-	return buf, nil
+// encodeVarInt returns the encoded unsigned integer into a Base128 variable-length integer.
+func encodeVarInt(value uint64) []byte {
+	var buf [10]byte
```
robsdedude:
```suggestion
	var buf [binary.MaxVarintLen64]byte
```

neo4j/internal/bolt/connect.go
```diff
@@ -260,55 +261,41 @@ func sendHandshakeConfirmation(ctx context.Context, conn io.ReadWriteCloser, bol
...
-		// The most significant bit is the continuation flag.
-		if b&0x80 == 0 {
-			break
-		}
-		shift += 7
-		if shift >= 64 {
-			return 0, fmt.Errorf("varint too long")
+		// If the continuation bit is not set, we've reached the end of the varint.
+		if buf[i]&0x80 == 0 {
+			value, n := binary.Uvarint(buf[:i+1])
```
robsdedude, line 285: Nice find :rocket: I didn't know that Go ships with these hand-dandy little helper functions.

neo4j/internal/bolt/connect_test.go
```diff
@@ -186,3 +189,159 @@ func TestPerformManifestNegotiationNoSupportedVersion(t *testing.T) {
...
+	tests := []struct {
+		value    uint64
+		expected []byte
+	}{
+		{0, []byte{0x00}},
+		{1, []byte{0x01}},
+		{127, []byte{0x7F}},
+		{128, []byte{0x80, 0x01}},
+		{300, []byte{0xAC, 0x02}},
+		{16384, []byte{0x80, 0x80, 0x01}},
```
robsdedude, line 235:
```suggestion
		{16384, []byte{0x80, 0x80, 0x01}},
		{^uint64(0), []byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01}},
```
Always a good idea to test the edge cases where possible :)

neo4j/internal/bolt/connect_test.go
```diff
@@ -186,3 +189,159 @@ func TestPerformManifestNegotiationNoSupportedVersion(t *testing.T) {
...
+	if err == nil {
+		t.Error("expected error for varint too long, got nil")
+	} else if err.Error() != "varint too long" {
+		t.Errorf("expected error 'varint too long', got %v", err)
+	}
+}
+
+// TestReadVarIntReadError verifies that a read error from the underlying reader is returned.
+func TestReadVarIntReadError(t *testing.T) {
+	reader := &errorRacingReader{}
```
robsdedude, line 300:
```suggestion
	t.Parallel()
	reader := &errorRacingReader{}
```
Consider adding this to all new tests that don't alter global state.

robsdedude (approved):
🤝 only small comments, no blockers

neo4j/internal/bolt/connect.go
StephenCathcart: Same here, I don't often peruse the binary package 😂

neo4j/internal/bolt/connect.go
StephenCathcart: I think we need the safeguard - even though we limit the buffer to 10 bytes and stop at the first byte without the continuation bit, it's still possible for the varint to be malformed - like such as encoding a value that doesn't fit in a uint64? I've added a unit test in the last commit to simulate this invalid scenario and verify that we correctly return the "failed to decode varint" error.

robsdedude: Absolutely! My bad. I was thinking that u64 goes perfectly into a 10 byte varint, but no, you're right. There are 6 spare bits that must be 0.

PR #620, opened 2025-02-10
Support google/uuid.UUID type as query parameter

jordan-bonecutter:
This type is pretty common and would be nice to support.

Happy to write tests if this is something that could be merged

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module github.com/neo4j/neo4j-go-driver/v5
 
 go 1.18
+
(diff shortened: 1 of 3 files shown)


StephenCathcart:
5.0 only takes bug fixes now, not new parameter types.

Already covered on 6.x, dbtype.UUID works as a query parameter and converts directly to/from google/uuid.UUID. On 5.0 you could pass it as a string or []byte for now.

Closing here.

PR #624, opened 2025-02-17
TestKit: skip one of the newly added summary tests

robsdedude:
Skips one of the newly introduced tests from
 * https://github.com/neo4j-drivers/testkit/pull/653


StephenCathcart (approved):
👍

PR #625, opened 2025-02-17
Fix: Use json.Marshal to create cache key

StephenCathcart:


StephenCathcart:
Experimental branch for cache key collisions

PR #626, opened 2025-02-25
Fix panic when calling driver.Close concurrently

robsdedude:
Doing so might well still yield unexpected or undesired results, but at least it shouldn't cause a panic.

Depends on:
 * https://github.com/neo4j/neo4j-go-driver/pull/628


StephenCathcart (approved):
Logic and tests look good 👍

AndyHeap-NeoTech (approved):
Tests and implementation look good.

StephenCathcart (approved):
Re-review approval 👍

PR #627, opened 2025-02-25
[4.4] Fix panic when calling driver.Close concurrently

robsdedude:
Doing so might well still yield unexpected or undesired results, but at least it shouldn't cause a panic.

Backport of https://github.com/neo4j/neo4j-go-driver/pull/626

Depends on:
 * https://github.com/neo4j/neo4j-go-driver/pull/629

### DIFF
--- a/neo4j/db/connection.go
+++ b/neo4j/db/connection.go
@@ -79,8 +79,6 @@ type Connection interface {
 	Bookmark() string
 	// Returns name of the remote server
 	ServerName() string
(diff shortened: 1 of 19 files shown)


StephenCathcart (approved):
Fix looks good

AndyHeap-NeoTech (approved):
Looks good.

StephenCathcart (approved):
~~Fix looks good~~ as discussed offline, looking at a way to not extend `Connection` interface

PR #640, opened 2025-04-30
Fix context cancelled during clean-up GOODBYE

robsdedude:
* Make sure that the pool's asynchronous connection closure (`GOODBYE`) does not get cancelled when the acquisition of return of a connection that triggered the clean-up ends (and thus cancels the parent context).
 * Lower log level from error to debug on errors occurring during connection closure or when the connection is already dead.


StephenCathcart (approved):
🚢

PR #652, opened 2025-08-13
Deprecation of Single

StephenCathcart:
Added missing deprecation and a new context-aware version of `Single`.

- Added `SingleWithContext`.
- Deprecated `Single` with notice to use `SingleWithContext`.
- Follows the same pattern as other `*WithContext` result helpers (`SingleTWithContext`, `CollectTWithContext`, `CollectWithContext`).


robsdedude (approved):
:ring: :no_good:

PR #655, opened 2025-08-14
File renames and test cleanup

StephenCathcart:
Removed "with_context" from various files. Merged go1.23 specific `Result` tests with other result tests.


MaxAake (approved):
✅

PR #657, opened 2025-08-15
Moved TLS comment to Driver Example

StephenCathcart:


neo4j/driver_examples_test.go
```diff
@@ -138,6 +143,32 @@ func ExampleDriver_verifyAuthenticationDriverLevel() {
...
+		config.TlsConfig = &tls.Config{
+			RootCAs:    certPool,
+			MinVersion: tls.VersionTLS12,
+		}
+	})
+	handleError(err)
+	defer handleClose(ctx, driver)
+
+	// Use the driver as normal - it will now trust your self-signed certificates
+	result, err := ExecuteQuery(ctx, driver, "RETURN 1 as num", nil, EagerResultTransformer)
```
robsdedude:
```suggestion
	result, err := ExecuteQuery(ctx, driver, "RETURN 1 AS num", nil, EagerResultTransformer)
```
nit: Cypher style guide :art:

neo4j/driver_examples_test.go
```diff
@@ -138,6 +143,32 @@ func ExampleDriver_verifyAuthenticationDriverLevel() {
 	handleError(err)
 }
 
+func ExampleConfig_tlsSelfSignedCertificates() {
```
robsdedude:
```suggestion
func ExampleDriver_tlsSelfSignedCertificates() {
```
Config does not live in this package. So either rename the example like above or move it into the `config` package to be displayed next to `config.Config` (instead of next to `neo4j.Driver`).

robsdedude (changes requested):

PR #666, opened 2025-09-03
Short-circuit retries on context timeout

zolstein:
Check explicitly for context completion in State.Continue to prevent unnecessary work that will inevitably fail.

This also fixes a bug where ExecuteQuery will loop without sleeping until a condition is hit (like MaxTransactionRetryTime or MaxDeadConnections) because the error handling sets skipSleep = true.

### DIFF
--- a/neo4j/session_with_context_test.go
+++ b/neo4j/session_with_context_test.go
@@ -183,7 +183,6 @@ func TestSession(outer *testing.T) {
 
 		inner.Run("Fails fast on context timeout", func(t *testing.T) {
 			_, pool, sess := createSession()
-			sess.driverConfig.MaxTransactionRetryTime = 10 * time.Second
 			pool.BorrowConn = &ConnFake{Alive: false}
 			numRetries := 0
 			ctx := context.Background()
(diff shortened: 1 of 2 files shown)


neo4j/session_with_context_test.go
```diff
@@ -181,6 +181,26 @@ func TestSession(outer *testing.T) {
...
 
+		inner.Run("Fails fast on context timeout", func(t *testing.T) {
+			_, pool, sess := createSession()
+			sess.driverConfig.MaxTransactionRetryTime = 10 * time.Second
+			pool.BorrowConn = &ConnFake{Alive: false}
+			numRetries := 0
+			ctx := context.Background()
+			ctx, cancel := context.WithTimeout(ctx, 0)
+			defer cancel()
+			<-ctx.Done()
```
zolstein, line 191: This unusual method of cancelling the context causes ctx.Err() to return "deadline exceeded", which represents a worst-case scenario - without the change to Continue, it will process the error as retryable and loop until it hits the MaxDeadConnections limit. By comparison, just cancelling the context the usual way will fail after one attempt, because the "context cancelled" error isn't retryable.

StephenCathcart: Nice explanation, great spot 👍

StephenCathcart:
Looks good, I'll port this to `6.x` too, thank you for the contribution, it's much appreciated :)

PR #668, opened 2025-09-15
Fix Close deadlock

StephenCathcart:
Potential fix for deadlocking when calling `Driver.Close` multiple times:

```go
for i := 0; i < 10; i++ {
    err = driver.Close(ctx)
    // can deadlock
}
```

- `Close` didn't unlock the mutex when the pool was already closed.
- First `Close` sets `pool = nil`; second `Close` sees `pool = nil` but doesn't unlock mutex.
- Third `Close` tries to acquire a mutex that's still locked by the second call.


MaxAake (approved):
✅

PR #669, opened 2025-09-16
[5.0] Fix Close deadlock

StephenCathcart:
Port of #668 - fix for deadlocking when calling `Driver.Close` multiple times.


MaxAake (approved):
✅

PR #675, opened 2025-10-07
Make PoolOutOfServers error retryable.

StephenCathcart:
When multiple goroutines access the pool concurrently, one may remove all servers from the routing table while another is trying to borrow a connection, resulting in a `PoolOutOfServers` error. This error should trigger a retry with a fresh routing table fetch, similar to `PoolTimeout`.

- Wrapped `PoolOutOfServers` as `ConnectivityError` for consistency.

### DIFF
--- a/neo4j/internal/retry/state.go
+++ b/neo4j/internal/retry/state.go
@@ -141,9 +141,6 @@ func IsRetryable(err error) bool {
 	if _, ok := err.(*errorutil.PoolTimeout); ok {
 		return true
 	}
-	if _, ok := err.(*errorutil.PoolOutOfServers); ok {
-		return true
-	}
 	var dbError *db.Neo4jError
 	if !errors.As(err, &dbError) {
 		return false
--- a/neo4j/internal/retry/state.go
+++ b/neo4j/internal/retry/state.go
@@ -141,6 +141,9 @@ func IsRetryable(err error) bool {
 	if _, ok := err.(*errorutil.PoolTimeout); ok {
 		return true
 	}
+	if _, ok := err.(*errorutil.PoolOutOfServers); ok {
+		return true
+	}
 	var dbError *db.Neo4jError
 	if !errors.As(err, &dbError) {
 		return false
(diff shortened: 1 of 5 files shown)


neo4j/internal/retry/state.go
```diff
@@ -141,6 +141,9 @@ func IsRetryable(err error) bool {
 	if _, ok := err.(*errorutil.PoolTimeout); ok {
 		return true
 	}
+	if _, ok := err.(*errorutil.PoolOutOfServers); ok {
+		return true
+	}
```
robsdedude: If my search-foo didn't let me down, I don't think that `IsRetryable` is ever being called on an unwrapped error. Therefore, I doubt this code-path ever gets hit.

If, however, we want `IsRetryable` to also work for unwrapped errors, there'd be more to do here. Many errors get wrapped in `ConnectivityError` which makes them retryable. So all of those would probably also need a special check here.

Not a blocker though, as I doubt it'll break anything. Except for devs after us wondering why this if was put here with some intention :upside_down_face:

StephenCathcart: Yeah, good catch, in my head, `IsRetryable` is a public API, therefore I thought we needed a case for the error for users with custom retry logic, but the internal retry flow in `Continue()` 
wraps errors before checking (`IsRetryable(errorutil.WrapError(lastErr))`), so the direct check never gets hit there.

Happy to remove this, especially if there needs to be more work anyway to make other unwrapped errors work correctly - what's your thoughts?

robsdedude: > therefore I thought we needed a case for the error for users with custom retry logic

Will such user-written retry loop see the wrapped or unwarpped error? If it ever sees the unwrapped/raw error (`PoolOutOfServers`), then I totally agree, that this check should remain (and other errors might also need to be added here).

StephenCathcart: Yeah, you're right, a user never sees unwrapped `PoolOutOfServers` errors - every path through `session.getConnection()`, `session.Run()`, etc. wraps them. 

I've removed that code path, added a simple test to capture the logic in `WrapError`, and added a couple of comments for future authors. I'll add a card to our backlog for investigating if the `IsRetryable` function needs a refactor for handling unwrapped errors.

PR #678, opened 2025-10-20
Remove retries when a connection acquisition timeout error is triggered

StephenCathcart:
Unification task to align with other Neo4j drivers.


StephenCathcart:
[[Go] Remove retries when an acquisition timeout error is triggered](https://trello.com/c/t7kTQqXo/6454-go-remove-retries-when-an-acquisition-timeout-error-is-triggered)

MaxAake (approved):
✅

PR #679, opened 2025-10-21
ADR 034: Unify Vector String representation

StephenCathcart:
Closes DRIVERS-49

### DIFF
--- a/neo4j/dbtype/vector.go
+++ b/neo4j/dbtype/vector.go
@@ -88,9 +88,6 @@ func formatElement[T VectorElement](element T) string {
 }
 
 func formatFloat(f float64, bitSize int) string {
-	if math.IsNaN(f) {
-		return "NaN"
-	}
 	if math.IsInf(f, 1) {
 		return "Infinity"
 	}
--- a/neo4j/dbtype/vector.go
+++ b/neo4j/dbtype/vector.go
@@ -77,9 +77,9 @@ func formatVectorData[T VectorElement](v Vector[T]) string {
 func formatElement[T VectorElement](element T) string {
 	switch e := any(element).(type) {
 	case float32:
-		return formatFloat32(e)
+		return formatFloat(float64(e), 32)
 	case float64:
-		return formatFloat64(e)
+		return formatFloat(e, 64)
 	case int8, int16, int32, int64:
 		return fmt.Sprintf("%d", e)
 	default:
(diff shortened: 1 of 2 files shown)


neo4j/dbtype/vector.go
```diff
@@ -17,10 +17,94 @@
...
+	for i, element := range v {
+		parts[i] = formatElement(element)
+	}
+	return strings.Join(parts, ", ")
+}
+
+func formatElement[T VectorElement](element T) string {
+	switch e := any(element).(type) {
+	case float32:
+		return formatFloat(float64(e))
```
robsdedude: Formatting a float32 as float64 is not ideal.

Try these cases in `TestVectorString`
```go
		{"cool", Vector[float64]{0.123}, "vector([0.123], 1, FLOAT NOT NULL)"},
		{"not cool", Vector[float32]{0.123}, "vector([0.123], 1, FLOAT32 NOT NULL)"},
```

The precision of the float type is used by the built-in stringer implementation to know where to cut off when turning binary into decimal places. Of course, there will always be imprecision by the finite nature of computers, but they're much smaller if the right type is being used.

Besides, I dislike the fact that Go forces use to cast to float64 first—wasted CPU cycles—but maybe negligible given that we're allocating a string anyway.

neo4j/dbtype/vector.go
```diff
@@ -17,10 +17,94 @@
...
+	if isWholeNumber(f) {
+		// Ensure we show at least one decimal place
+		return strconv.FormatFloat(f, 'f', 1, 64)
+	}
+	return strconv.FormatFloat(f, 'g', -1, 64)
+}
+
+func isWholeNumber(f float64) bool {
+	const epsilon = 1e-9
+	return math.Abs(f-math.Trunc(f)) < epsilon
```
robsdedude: I don't think this is ideal.

```go
{"oh noes", Vector[float64]{1e-50}, "vector([1e-50], 1, FLOAT NOT NULL)"},
```

It shouldn't pretend small floats are zero when they're not in reality.

Are there cases where this wouldn't work?

 ```suggestion
	return f == math.Trunc(f)
```

File: neo4j/dbtype/vectortypes_test.go [hunk 3]
```diff
@@ -167,3 +169,59 @@ func TestVectorEmptyAndNil(t *testing.T) {
 		testutil.AssertDeepEquals(t, nilVec[0], 1.0)
 	})
 }
+
+func TestVectorString(t *testing.T) {
```
robsdedude, line 173: Nit: could add more test cases:
 * sub-normal floats
 * NaNs with payload
 * signaling/non-signaling NaNs

robsdedude (changes requested):

neo4j/dbtype/vector.go
StephenCathcart: Cheers, went around in a loop thinking `return f == math.Trunc(f)` might have some caveat, but can't find one. Happy to go with this for sure 👍

neo4j/dbtype/vectortypes_test.go
StephenCathcart: I've added a couple more `sub-normal` float tests and `precision` tests. I don't think Go exposes different NaN types (signaling NaN) though `NaN()` (which is a quiet NaN) - this is covered in the test case of "special floats".

neo4j/dbtype/vector.go
StephenCathcart: I've separated out the logic slightly and can confirm the test cases pass (see `// Precision test cases`)

robsdedude (approved):

PR #680, opened 2025-10-21
Remove deprecations from Neo4jError

StephenCathcart:
More work is required internally before these can be deprecated with appropriate replacements.


MaxAake (approved):
✅

PR #681, opened 2025-10-22
Deprecate StatementType in favor of QueryType for Driver unification

StephenCathcart:
Closes DRIVERS-55

- Added `QueryType` as alias for `StatementType` in both `neo4j` and `db` packages.
- Added deprecation notices to `StatementType` and all related constants.
- Updated `ResultSummary` interface to include `QueryType()`.
- Updated internals to use `QueryType` constants.


AndyHeap-NeoTech (approved):
Looks good to me

PR #683, opened 2025-10-24
Update deprecation comments for Id, StartId and EndId

StephenCathcart:
Closes DRIVERS-58

Driver states that `Id`, `StartId` and `EndId` WILL be removed in 6.0 which is no longer true. Update to remove this and update lint rules to be more generic.


AndyHeap-NeoTech (approved):
LGTM

PR #686, opened 2025-11-24
Refactor Vector implementation

StephenCathcart:
Closes DRIVERS-115

- `Vector` is now a struct that holds a slice of elements.
- Tightened constraint on `VectorElement`. 
- Added `Vector` type variants to `RecordValue` and `PropertyValue` interfaces, enabling extraction via `GetRecordValue` and `GetProperty`.
- Aliases are working as intended. User-defined types are unsupported (similar to other dbtype structures).

```go
vec := neo4j.Vector[float64]{Elems: []float64{1.0, 2.0, 3.0}}

// Create a node with a vector property
result, err := neo4j.ExecuteQuery(ctx, driver,
	"CREATE (n:VectorExample {vec: $vec}) RETURN n, n.vec AS vec",
	map[string]any{"vec": vec},
	neo4j.EagerResultTransformer)

record := result.Records[0]

// Use GetRecordValue to extract vector from record
recordVec, _, err := neo4j.GetRecordValue[neo4j.Vector[float64]](record, "vec")

// Use GetProperty to extract vector from node
node := record.Values[0].(neo4j.Node)
propVec, err := neo4j.GetProperty[neo4j.Vector[float64]](node, "vec")
```

### DIFF
--- a/neo4j/dbtype/vector.go
+++ b/neo4j/dbtype/vector.go
@@ -35,7 +35,7 @@ type Vector[T VectorElement] struct {
 }
 
 // String returns the string representation of this Vector in the format:
-// vector([data], length, type NOT NULL).
+// vector([data], length, type NOT NULL)
 func (v Vector[T]) String() string {
 	dataStr := formatVectorData(v.Elems)
 	length := len(v.Elems)
--- a/neo4j/dbtype/vector.go
+++ b/neo4j/dbtype/vector.go
@@ -26,17 +26,58 @@ import (
 
 // VectorElement represents the supported element types for Vector.
 type VectorElement interface {
-	~float64 | ~float32 | ~int8 | ~int16 | ~int32 | ~int64
+	float64 | float32 | int8 | int16 | int32 | int64
 }
 
 // Vector represents a fixed-length array of numeric values.
-type Vector[T VectorElement] []T
+type Vector[T VectorElement] interface {
+	// Len returns the number of elements in the vector.
+	Len() int
 
-// String returns the string representation of this Vector in the format:
-// vector([data], length, type NOT NULL)
-func (v Vector[T]) String() string {
-	dataStr := formatVectorData(v)
-	length := len(v)
+	// At returns the element at index i. It panics if i is out of range.
+	At(i int) T
+
+	// Slice returns the elements as a new slice. The returned slice is a copy.
+	Slice() []T
+
+	// String returns the string representation of this Vector in the format:
+	// vector([data], length, type NOT NULL).
+	String() string
+}
+
+type vector[T VectorElement] struct {
+	elems []T
+}
+
+// NewVector constructs a Vector from the provided elements.
+func NewVector[T VectorElement](elems ...T) Vector[T] {
+	cp := make([]T, len(elems))
+	copy(cp, elems)
+	return &vector[T]{elems: cp}
+}
+
+// NewVectorFromSlice constructs a Vector from a slice.
+func NewVectorFromSlice[T VectorElement](elems []T) Vector[T] {
+	return NewVector(elems...)
+}
+
+func (v *vector[T]) Len() int {
+	return len(v.elems)
+}
+
+func (v *vector[T]) At(i int) T {
+	return v.elems[i]
+}
+
+func (v *vector[T]) Slice() []T {
+	cp := make([]T, len(v.elems))
+	copy(cp, v.elems)
+	return cp
+}
+
+func (v *vector[T]) String() string {
+	dataStr := formatVectorData(v.elems)
+	length := len(v.elems)
 	typeStr := getVectorTypeString[T]()
 
 	return fmt.Sprintf("vector([%s], %d, %s)", dataStr, length, typeStr)
@@ -62,7 +103,7 @@ func getVectorTypeString[T VectorElement]() string {
 	}
 }
 
-func formatVectorData[T VectorElement](v Vector[T]) string {
+func formatVectorData[T VectorElement](v []T) string {
 	if len(v) == 0 {
 		return ""
 	}
--- a/neo4j/dbtype/vector.go
+++ b/neo4j/dbtype/vector.go
@@ -30,54 +30,15 @@ type VectorElement interface {
 }
 
 // Vector represents a fixed-length array of numeric values.
-type Vector[T VectorElement] interface {
-	// Len returns the number of elements in the vector.
-	Len() int
-
-	// At returns the element at index i. It panics if i is out of range.
-	At(i int) T
-
-	// Slice returns the elements as a new slice. The returned slice is a copy.
-	Slice() []T
-
-	// String returns the string representation of this Vector in the format:
-	// vector([data], length, type NOT NULL).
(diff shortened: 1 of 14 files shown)


neo4j/dbtype/vector.go
```diff
@@ -26,17 +26,19 @@ import (
...
 
 // Vector represents a fixed-length array of numeric values.
-type Vector[T VectorElement] []T
+type Vector[T VectorElement] struct {
+	Elems []T
+}
 
 // String returns the string representation of this Vector in the format:
-// vector([data], length, type NOT NULL)
+// vector([data], length, type NOT NULL).
```
stefano-ottolenghi: The trailing dot is misleading, as it could be interpreted as being part of the string representation.

neo4j/dbtype/vector_example_test.go
```diff
@@ -23,41 +23,44 @@ import (
...
+	recordVec, _, err := neo4j.GetRecordValue[neo4j.Vector[float64]](record, "vec")
 	if err != nil {
 		panic(err)
 	}
 
-	if v, ok := result.Records[0].Values[0].(dbtype.Vector[float64]); ok {
-		fmt.Printf("Read vector: %v\n", v)
+	// Use GetProperty to extract vector from node
+	node := record.Values[0].(neo4j.Node)
+	propVec, err := neo4j.GetProperty[neo4j.Vector[float64]](node, "vec")
```
stefano-ottolenghi, line 64: Back in the days, I put this note https://neo4j.com/docs/go-manual/current/data-types/ in the manual:

> ... cast the value to string before using it as a string: `name := record.AsMap()["name"].(string)`

You use a different way of extracting values from records - is there one way preferred over the other and, if yes, why?

neo4j/dbtype/vector.go
StephenCathcart: Yep agree 👍

neo4j/dbtype/vector_example_test.go
StephenCathcart: Nice observation - you can grab values directly when you trust the `record` shape and want minimum overhead, but the `GetRecordValue`/`GetProperty` helpers give type-safe errors and nil detection, so I recommend them for any code that handles unexpected data or needs clearer failure states. I'll update the example.

neo4j/record.go
```diff
@@ -27,7 +27,8 @@ type RecordValue interface {
 		Point2D | Point3D |
 		Date | LocalTime | LocalDateTime | Time | Duration | time.Time | /* OffsetTime == Time == dbtype.Time */
 		[]byte | []any | map[string]any |
-		Node | Relationship | Path
+		Node | Relationship | Path |
+		Vector[int8] | Vector[int16] | Vector[int32] | Vector[int64] | Vector[float32] | Vector[float64]
```
robsdedude, line 31: Nit: could expand `GetRecordValue` tests to cover new types.

neo4j/dbtype/vectortypes_test.go
```diff
@@ -20,154 +20,21 @@ package dbtype
...
-		nilVec = append(nilVec, 1.0)
-		testutil.AssertLen(t, nilVec, 1)
-		testutil.AssertDeepEquals(t, nilVec[0], 1.0)
-	})
+	var _ Vector[float64]
+	var _ Vector[float32]
+	var _ Vector[int8]
+	var _ Vector[int16]
+	var _ Vector[int32]
+	var _ Vector[int64]
```
robsdedude, line 37: nit: might want to add `dbtype` aliases

neo4j/dbtype/vectortypes_test.go
```diff
@@ -179,50 +46,50 @@ func TestVectorString(t *testing.T) {
 		expected string
 	}{
 		// Empty vectors
-		{"empty int8", Vector[int8]{}, "vector([], 0, INTEGER8 NOT NULL)"},
-		{"empty int16", Vector[int16]{}, "vector([], 0, INTEGER16 NOT NULL)"},
-		{"empty int32", Vector[int32]{}, "vector([], 0, INTEGER32 NOT NULL)"},
-		{"empty int64", Vector[int64]{}, "vector([], 0, INTEGER NOT NULL)"},
-		{"empty float32", Vector[float32]{}, "vector([], 0, FLOAT32 NOT NULL)"},
-		{"empty float64", Vector[float64]{}, "vector([], 0, FLOAT NOT NULL)"},
+		{"empty int8", Vector[int8]{Elems: []int8{}}, "vector([], 0, INTEGER8 NOT NULL)"},
```
robsdedude, line 49: While looking at this, I'm ever so slightly annoyed by having to read (and potentially having to write) the inner vector type twice...

I mean users can avoid this by writing a simple
```
func makeVec[T dbtype.VectorElement](elems []T) neo4j.Vector[T] {
	return neo4j.Vector[T]{Elems: elems}
}
```
or similar... As a more seasoned Go dev, what's your gut feeling? Is it worth adding such a helper function to the driver or is it unnecessary noise?

neo4j/record.go
StephenCathcart: I've added some to `GetProperty` too 👍

neo4j/dbtype/vectortypes_test.go
StephenCathcart: These already implicitly call `dbtype.Vector` as they're located in the `neo4j/dbtype` package. I could add a new test file, `neo4j/aliases_test.go`, and add something like the following in:

```
func TestVectorAlias(t *testing.T) {
	t.Parallel()

	var _ dbtype.Vector[float64] // either test compile of dbtype package
	var _ Vector[float32] // or test compile of implict neo4j package alias
        ....etc
}
```

Or remove the existing test from dbtype and test both in this new aliases_test file like above - thoughts?

neo4j/dbtype/vectortypes_test.go
StephenCathcart: I'd skip it for now. We use struct literals for `Point2D`/`Point3D`, and helpers seem to be for complex cases like `DurationOf`. The current syntax is explicit, but I think that's the Go way, and I'd be guessing on API design by adding the helper or its use.

Should we open a GitHub discussion to see what users think of the Vector API in general? Then we can easily add a helper in a later minor if needed?

neo4j/dbtype/vectortypes_test.go
robsdedude: I should've seen that :facepalm:

In such case, it's probably not worth the hassle. I'd avoid having similar tests spread across files or even packages like that.

neo4j/dbtype/vectortypes_test.go
robsdedude: In such case, I think this is fine as is :ok_hand:

I'll leave it up to you whether you want to start a discussion. We usually only do it for preview & experimental features as those allow us to actually break the API if we get feedback that makes us want to do so. But vectors will be stable right away. So we could only do additive changes (like the construction function).

robsdedude (approved):
:arrow_upper_right:

PR #687, opened 2026-03-10
ProfiledPlan: expose existence for all metrics

robsdedude:
Deprecate `ResultSummary.Profile()` returning a `ProfiledPlan` in favor of `ResultSummary.QueryProfile()` returning a `Profile`.

The main difference between the profile interfaces is that the new one exposes whether stats (such as `DbHits()`, `Time()`, etc.) have been recorded by returning `(T, bool)` instead of `T`. Further, `ProfiledPlan.Records()` corresponds to the new `Profile.Rows()` to better align with Neo4j's embedded API, Bolt, and the raw arguments attached to the profile steps.

According to a chat with the Cypher runtime team, the absence of the value is not equivalent to it being zero. Therefore it makes sense to express this information in the API. The information is encoded in the bolt protocol through absence or existence of the respective fields. The driver logic has been changed to use this information instead of checking for fallback zero values.

Closes: DRIVERS-284

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/705


robsdedude:
Please don't merge yet. I'd like to wait for this work to be ready for merging across all drivers first.

StephenCathcart (approved):
👍 💯

robsdedude:
@StephenCathcart ready for a final review and merging when you're happy with it.

PR #693, opened 2026-04-27
Add idempotent one-shot retry to session.Run (ADR-038)

StephenCathcart:
Auto-commit transactions now retry once when the server returns an idempotent FAILURE on RUN. Adds DisableAutoCommitRetries config option at driver and session level (session-level *bool overrides driver). Bolt Telemetry contract is refined so the onSuccess callback fires whenever the telemetry intent is resolved - server SUCCESS or skipped because the protocol/server doesn't carry it - letting session.Run cleanly gate retry on whether the FAILURE came from RUN vs TELEMETRY.

Closes: DRIVERS-312

### DIFF
--- a/neo4j/db/errors.go
+++ b/neo4j/db/errors.go
@@ -150,6 +150,14 @@ func (e *Neo4jError) MarkRetriable() {
 	e.retriable = true
 }
 
+// IsIdempotent reports whether the server marked this error as idempotent in
+// its diagnostic record, meaning the error is guaranteed not to have altered
+// the state of any database. Defaults to false when the field is absent.
+func (e *Neo4jError) IsIdempotent() bool {
+	v, ok := e.GqlDiagnosticRecord["_idempotent"].(bool)
+	return ok && v
+}
+
 // ContainsGqlStatus returns whether there is an error with the given GQLSTATUS in this GQL error chain,
 // beginning the search from this exception.
 func (e *Neo4jError) ContainsGqlStatus(gqlStatus string) bool {
--- a/neo4j/db/errors.go
+++ b/neo4j/db/errors.go
@@ -150,14 +150,6 @@ func (e *Neo4jError) MarkRetriable() {
 	e.retriable = true
 }
 
-// IsIdempotent reports whether the server marked this error as idempotent in
-// its diagnostic record, meaning the error is guaranteed not to have altered
-// the state of any database. Defaults to false when the field is absent.
-func (e *Neo4jError) IsIdempotent() bool {
-	v, ok := e.GqlDiagnosticRecord["_idempotent"].(bool)
-	return ok && v
-}
-
 // ContainsGqlStatus returns whether there is an error with the given GQLSTATUS in this GQL error chain,
 // beginning the search from this exception.
 func (e *Neo4jError) ContainsGqlStatus(gqlStatus string) bool {
--- a/neo4j/session.go
+++ b/neo4j/session.go
@@ -791,12 +791,10 @@ func (s *session) Run(ctx context.Context,
 		)
 		if err != nil {
 			s.pool.Return(ctx, conn)
-			var neo4jErr *db.Neo4jError
 			if !disableAutoCommitRetries &&
 				attempt == 0 &&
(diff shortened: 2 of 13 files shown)


neo4j/db/errors.go
```diff
@@ -150,6 +150,14 @@ func (e *Neo4jError) MarkRetriable() {
 	e.retriable = true
 }
 
+// IsIdempotent reports whether the server marked this error as idempotent in
+// its diagnostic record, meaning the error is guaranteed not to have altered
+// the state of any database. Defaults to false when the field is absent.
+func (e *Neo4jError) IsIdempotent() bool {
```
robsdedude: I think no other driver added such an endpoint to its public API. The closes to this is probably the JS PR (sill WiP), which adds the diagnostic record to the public DiagnosticRecord type (which is more of a type promise than a hard type guarantee): https://github.com/neo4j/neo4j-javascript-driver/pull/1404/changes#diff-f1b553a9f219980025c569c7a2dacb0c8212673a5bedd7b32a75c1074ea16434R18

neo4j/session_test.go
```diff
@@ -508,6 +508,139 @@ func TestSession(outer *testing.T) {
...
+				GqlDiagnosticRecord: map[string]any{"_idempotent": true},
+			}
+		}
+		newOtherErr := func() *db.Neo4jError {
+			return &db.Neo4jError{
+				Code: "Neo.ClientError.MadeUp.Other",
+				Msg:  "different",
+			}
+		}
+		mkSession := func(conf config.Config, sessConfig SessionConfig) (*PoolFake, *session) {
```
robsdedude: Nit: this isn't C 😜 
```suggestion
		makeSession := func(conf config.Config, sessConfig SessionConfig) (*PoolFake, *session) {
```
(requires more edits, won't comment in other places to reduce noise)

neo4j/session_test.go
```diff
@@ -508,6 +508,139 @@ func TestSession(outer *testing.T) {
...
+
+		inner.Run("propagates second-attempt error", func(t *testing.T) {
+			pool, sess := mkSession(config.Config{MaxConnectionPoolSize: 100}, SessionConfig{})
+			attempts := 0
+			pool.BorrowHook = func() (idb.Connection, error) {
+				attempts++
+				if attempts == 1 {
+					return &ConnFake{Alive: true, RunErr: newIdempotentErr()}, nil
+				}
+				return &ConnFake{Alive: true, RunErr: newOtherErr()}, nil
```
robsdedude, line 568:
```suggestion
				return &ConnFake{Alive: true, RunErr: newIdempotentErr()}, nil
```
I suggest to make the 2nd error also idempotent because even then the driver shouldn't retry. Maybe it's even worthwhile to have both cases tested.

neo4j/db/errors.go
StephenCathcart: agreed, moved to an unexported helper in session.go 👍

neo4j/session_test.go
StephenCathcart: added a sub-test for both-idempotent, kept the existing one for cross-type 👍

neo4j/session_test.go
robsdedude, line 1: Optional: could maybe sprinkle some more `t.Parallel()` in there.

neo4j/session.go
```diff
@@ -728,60 +744,92 @@ func (s *session) Run(ctx context.Context,
...
+					DisCats: s.config.NotificationsDisabledCategories,
+					DisClas: s.config.NotificationsDisabledClassifications,
+				},
+			},
+		)
+		if err != nil {
+			s.pool.Return(ctx, conn)
+			if !disableAutoCommitRetries &&
+				attempt == 0 &&
+				telemetryResolved &&
```
robsdedude, line 796: For the record: we had a chat about this because we think it's potentially fragile. If we ever add another message that's pipeline before the `RUN`, this heuristic might fail to achieve its intended purpose. However, we couldn't come up with a cleaner way without major refactoring of the driver.

robsdedude (approved):

PR #695, opened 2026-05-14
Add object mapping for struct query parameters (preview)

StephenCathcart:
Structs passed as Cypher parameter values are now packed via reflection over exported fields, keyed by field name or the `neo4j:"name,omitempty"` tag. encoding/json semantics for the tag: "-" skips, omitempty drops zero values, anonymous embedded structs flatten. Driver-known types (time.Time, dbtype.*) keep their existing on-wire encoding; field values bolt cannot serialise still return UnsupportedTypeError.

Closes: DRIVERS-104


neo4j/test-integration/types_test.go
```diff
@@ -420,7 +420,9 @@ func TestTypes(outer *testing.T) {
 	})
 
 	outer.Run("Un-convertible Go types", func(inner *testing.T) {
+		// Structs are now packable; chan keeps this un-convertible.
```
fbiville: question: out of curiosity mostly, if the `chan` field is explicitly omitted, does the packing work?

neo4j/internal/mapping/mapping.go
```diff
@@ -0,0 +1,120 @@
...
+	typeCache.Store(t, fields)
+	return fields
+}
+
+func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
+	var out []fieldInfo
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		tag := f.Tag.Get(tagName)
+		idx := append(append([]int(nil), indexPath...), i)
```
MaxAake, line 100:
```suggestion
		idx := append(indexPath, i)
```
Might be me failing at go knowledge, but couldn't this line be simplified like this if the places this is called where `indexPath` is `nil` are altered to instead pass in an empty `[]int`?

So line 73 would change to `buildFields(t, []int(nil))`

neo4j/driver.go
```diff
@@ -455,6 +455,12 @@ func (d *driver) VerifyAuthentication(ctx context.Context, auth *AuthToken) (err
 //
 //	result, err := ExecuteQuery[*EagerResult](ctx, driver, query, params, EagerResultTransformer)
 //
+// Parameter mapping is part of the Object Mapping preview feature (see README on what it means in
+// terms of support and compatibility guarantees). When a value in params is a user-defined struct
+// (or pointer to one), its exported fields are sent as a Cypher map. The struct tag
+// `neo4j:"name,omitempty"` renames a field, "-" skips it, and "omitempty" drops zero values.
```
MaxAake: This line is a little unclear to me.

```suggestion
// `neo4j:"orange"` renames the tagged field to `orange,` `neo4j:"-"` skips it, and `neo4j:",omitempty"` (note the comma) drops zero values from the field rather than transmitting it. renaming and dropping zero values can be combined with a tag like `neo4j:"blue,omitempty"`, which will both rename the field to `blue` and drop zero values
```

This may not be perfect, but it shows the extra information I think would be useful

neo4j/internal/bolt/outgoing_test.go
```diff
@@ -1091,6 +1106,125 @@ func TestOutgoing(ot *testing.T) {
...
+				"e": struct {
+					Inner movie
+				}{Inner: movie{Title: "Heat", Released: 1995}},
+			},
+			expect: map[string]any{
+				"e": map[string]any{
+					"Inner": map[string]any{"Title": "Heat", "Released": int64(1995)},
+				},
+			},
+		},
```
MaxAake, line 1252: I think it would be interesting to include a test for an inner struct that has tags applied to ensure they work.

MaxAake (commented):
💬

neo4j/test-integration/types_test.go
StephenCathcart: yeah. tag it `neo4j:"-"` (or unexport it) and it drops before packing. added a test for it 👍

neo4j/internal/mapping/mapping.go
StephenCathcart: Need the copy here. `append(indexPath, i)` can reuse the same backing array, so without copying a later fields path can mess with an earlier one. `encoding/json` copies in typeFields for the same reason 👍

neo4j/driver.go
StephenCathcart: reworded: split the three forms out and called out the leading comma 👍

neo4j/internal/bolt/outgoing_test.go
StephenCathcart: added a case with rename/skip/omitempty tags on the inner struct 👍

neo4j/internal/mapping/mapping.go
MaxAake: Ahhhh, yeah that does makes sense! 👍🏻

neo4j/internal/bolt/outgoing.go
```diff
@@ -380,7 +381,16 @@ func (o *outgoing) packStruct(x any) {
 	case dbtype.Vector[float64]:
 		o.packer.VectorFloat64(v.Elems)
 	default:
-		o.onPackErr(&db.UnsupportedTypeError{Type: reflect.TypeOf(x)})
+		m, err := mapping.StructAsMap(x)
```
robsdedude: Nit: `StructAsMap` potentially returns an error value. If so, the value of `err` is just used for a `nil`-check and then discarded; this seems to be wasted effort (granted not on the happy/hot path, so no big deal, really).

neo4j/internal/bolt/outgoing_test.go
```diff
@@ -488,6 +488,21 @@ func TestOutgoing(ot *testing.T) {
...
+	type tagged struct {
+		Title string `neo4j:"title"`
+		Skip  string `neo4j:"-"`
+		Empty string `neo4j:",omitempty"`
+	}
+	type base struct{ ID string }
+	type embedded struct {
+		base
+		Name string
+	}
```
robsdedude, line 507: Optional corner case to test:

```go
	type empty struct{}
	// ...
		{
			name: "user struct: without fields",
			inp: map[string]any{
				"e": empty{},
			},
			expect: map[string]any{
				"e": map[string]any{},
			},
		},
		{
			name: "user struct: point without fields",
			inp: map[string]any{
				"e": &empty{},
			},
			expect: map[string]any{
				"e": map[string]any{},
			},
		},
		{
			name: "user struct: nil point without fields",
			inp: map[string]any{
				"e": (*empty)(nil),
			},
			expect: map[string]any{
				"e": nil,
			},
		},
```
(tested it and it works)

neo4j/test-integration/types_test.go
robsdedude: Nit: the comment itself makes sense in the context of reading this PR. But imagine coming across this line in 5 years. “What ‘now’?”. I think this comment should go 🗑️

neo4j/internal/mapping/mapping.go
```diff
@@ -0,0 +1,120 @@
...
+	name      string
+	index     []int
+	omitEmpty bool
+}
+
+var typeCache sync.Map // map[reflect.Type][]fieldInfo
+
+// StructAsMap walks the exported fields of v (struct or pointer to struct)
+// and returns a map keyed by each field's Cypher property name. A nil pointer
+// returns (nil, nil); the caller maps that to bolt NULL.
```
robsdedude: *Nit picking intensifies* (please feel free to ignore this)

> the caller maps that to bolt NULL.

Why does the callee's comment concern itself with that the caller (of which more could be added anytime) does with the return value?

neo4j/internal/mapping/mapping.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	fields := fieldsOf(rv.Type())
+	out := make(map[string]any, len(fields))
+	for _, f := range fields {
+		fv := rv.FieldByIndex(f.index)
+		if f.omitEmpty && fv.IsZero() {
+			continue
+		}
+		out[f.name] = fv.Interface()
```
robsdedude, line 66: I don't know if it's worth the added complexity, but I strongly suspect (no tests conducted) that instead of building a `map[string]any` here (spending time on turning `reflect.Value`s into `any`s) and then calling `packMap` (back at the call site in `packX`) it'd be a fair bit faster to directly drive the packer from within the conversion logic and for instance call `packV(fv)` in a similar way it's done inside the `reflect.Map` case inside of `packX`.

The packing, after all, is the driver's hot path. Whether this feature in particular (packing custom user structs) should be considered hot, too, I don't know. And maybe you deem the much tighter coupling my suggested Design implies not worth the (to be confirmed) performance gain.

neo4j/internal/mapping/mapping.go
```diff
@@ -0,0 +1,120 @@
...
+
+const tagName = "neo4j"
+
+type fieldInfo struct {
+	name      string
+	index     []int
+	omitEmpty bool
+}
+
+var typeCache sync.Map // map[reflect.Type][]fieldInfo
```
robsdedude, line 37: I there risk of this cache growing unboundedly? E.g., by users throwing an unbounded amount of structs created at run-time (`reflect.StructOf`) into the driver? I really hope nobody would think this is a good idea but you know how software is. If there's a footgun somewhere sooner or later somebody will find it.

I might be overthinking this 🫥

neo4j/internal/mapping/mapping.go
```diff
@@ -0,0 +1,120 @@
...
+func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
+	var out []fieldInfo
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		tag := f.Tag.Get(tagName)
+		idx := append(append([]int(nil), indexPath...), i)
+
+		// Recurse into anonymous embeds - their inner fields stay reachable
+		// via FieldByIndex even when the wrapper type is unexported. Pointer
+		// kind skipped to avoid nil-deref on the walk.
```
robsdedude: Pointer embeds being ignored seems like a caveat worth documenting (or fixing).

neo4j/internal/mapping/mapping.go
```diff
@@ -0,0 +1,120 @@
...
+		name, omitEmpty := parseTag(tag)
+		if name == "" {
+			name = f.Name
+		}
+		out = append(out, fieldInfo{name: name, index: idx, omitEmpty: omitEmpty})
+	}
+	return out
+}
+
+func parseTag(tag string) (name string, omitEmpty bool) {
```
robsdedude, line 130: The logic ignores unknown tags. Is that common design in Go? What does `json` do, for instance? I'm thinking a warning or error to help the user detect typos might be helpful.

neo4j/test-integration/types_test.go
StephenCathcart: reworded to drop the "now"

neo4j/internal/mapping/mapping.go
StephenCathcart: kept it for now, direct-driving would tie mapping to the packer and this isn't a proven hot path. happy to benchmark a follow-up if it pays off given it's preview 👍 .

neo4j/internal/mapping/mapping.go
StephenCathcart: same as encoding/json's fieldCache, a sync.Map keyed by reflect.Type, so the same tradeoff as the stdlib I think. only really unbounded with runtime-generated types via reflect.StructOf. can add a note if you'd rather flag it though?

neo4j/internal/mapping/mapping.go
StephenCathcart: fixed, pointer embeds flatten like value embeds now and a nil embed skips its fields 👍

neo4j/internal/mapping/mapping.go
StephenCathcart: json does the same, it silently ignores unknown tag options. erroring at pack time can't tell a typo from a future option, so it felt too aggressive. could add a doc note of the supported options if that's better?

neo4j/internal/mapping/mapping.go
robsdedude: Frankly, if this approach it's good enough for the stdlib, I'm clearly overthinking this 😂

neo4j/internal/mapping/mapping.go
robsdedude: The options are already in the user-facing documentation. Just wanted to double check on the design. All good from my side 👍

robsdedude (approved):
🗺️

PR #696, opened 2026-05-15
Add UUID type for Bolt 6.1

StephenCathcart:
Closes: DRIVERS-362


neo4j/internal/bolt/bolt6_test.go
```diff
@@ -182,17 +186,18 @@ func TestBolt6(outer *testing.T) {
...
 				offerings: []protocolVersion{
-					{major: 6, minor: 1, back: 0},
-					{major: 6, minor: 0, back: 0},
+					{major: 6, minor: 1, back: 1},
 					{major: 5, minor: 8, back: 8},
 					{major: 4, minor: 4, back: 2},
 				},
-				description: "Server offers newer version (6.1) but driver selects supported version (6.0)",
+				expectedMinor: 1,
+				description:   "Server offers 6.1; driver picks 6.1 as the highest supported minor",
```
MaxAake, line 200: How about creating a new `"newer version offered but not selected"` with a version higher than Bolt 6.1?

MaxAake (approved):
only a tiny test suggestion, not a blocker

neo4j/internal/bolt/bolt6_test.go
StephenCathcart: added a 6.2-offered case that falls back to 6.1 👍

PR #703, opened 2026-07-16
Add object mapping for query results (preview)

StephenCathcart:
Closes: DRIVERS-103

Adds the output side of object mapping (preview): query results map into user structs, reversing the struct-parameter mapping from #695. 

Fields follow the same `neo4j` tags (`neo4j:"name"` renames, `neo4j:"-"` skips). A record with a single node, relationship, or map column maps that value's properties directly; otherwise columns map by name. Missing or null properties leave zero values, unmatched properties are ignored.

Entry points: 
- `As[T]` (one record)
- `CollectAs[T]` / `SingleAs[T]` (streamed `Result`)
- `CollectRecordsAs[T]` (`ExecuteQuery`'s `EagerResult.Records`)

## Usage

```go
type Movie struct {
    Title    string `neo4j:"title"`
    Released int64  `neo4j:"released"`
}

result, err := session.Run(ctx, "MATCH (m:Movie) RETURN m", nil)
if err != nil {
    return err
}
movies, err := neo4j.CollectAs[Movie](ctx, result)
```

## Before / after

**As** - one record:
```go
// before
node, _, err := neo4j.GetRecordValue[neo4j.Node](record, "m")
if err != nil {
    return err
}
movie := Movie{Title: node.Props["title"].(string), Released: node.Props["released"].(int64)}

// after
movie, err := neo4j.As[Movie](record)
```

**CollectAs** - every record in a Result:
```go
// before
movies, err := neo4j.CollectT(ctx, result, func(r *neo4j.Record) (Movie, error) {
    n, _, err := neo4j.GetRecordValue[neo4j.Node](r, "m")
    if err != nil {
        return Movie{}, err
    }
    return Movie{Title: n.Props["title"].(string), Released: n.Props["released"].(int64)}, nil
})

// after
movies, err := neo4j.CollectAs[Movie](ctx, result)
```

**SingleAs** - the single record in a Result:
```go
// before
movie, err := neo4j.SingleT(ctx, result, func(r *neo4j.Record) (Movie, error) {
    n, _, err := neo4j.GetRecordValue[neo4j.Node](r, "m")
    if err != nil {
        return Movie{}, err
    }
    return Movie{Title: n.Props["title"].(string), Released: n.Props["released"].(int64)}, nil
})

// after
movie, err := neo4j.SingleAs[Movie](ctx, result)
```

**CollectRecordsAs** - ExecuteQuery's eager records:
```go
// before
res, err := neo4j.ExecuteQuery(ctx, driver, cypher, params, neo4j.EagerResultTransformer)
if err != nil {
    return err
}
movies := make([]Movie, len(res.Records))
for i, r := range res.Records {
    n, _, err := neo4j.GetRecordValue[neo4j.Node](r, "m")
    if err != nil {
        return err
    }
    movies[i] = Movie{Title: n.Props["title"].(string), Released: n.Props["released"].(int64)}
}

// after
res, err := neo4j.ExecuteQuery(ctx, driver, cypher, params, neo4j.EagerResultTransformer)
if err != nil {
    return err
}
movies, err := neo4j.CollectRecordsAs[Movie](res.Records)
```

### DIFF
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -180,11 +180,11 @@ func assign(dst reflect.Value, raw any) error {
 	if raw == nil {
 		return nil
 	}
-	if dst.Kind() == reflect.Ptr {
+	for dst.Kind() == reflect.Ptr {
 		if dst.IsNil() {
 			dst.Set(reflect.New(dst.Type().Elem()))
 		}
-		return assign(dst.Elem(), raw)
+		dst = dst.Elem()
 	}
 	sv := reflect.ValueOf(raw)
 	dt := dst.Type()
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -69,8 +69,9 @@ func StructAsMap(v any) (map[string]any, bool) {
 	return out, true
 }
 
-// fieldByIndex walks index from v through pointer embeds. With alloc it allocates
-// nil embeds (false if one is unexported/unsettable); without, it stops false at the first nil.
+// fieldByIndex walks index from v to the target field, following pointer embeds.
+// With alloc it allocates nil embeds, returning false if one is unexported (and
+// so unsettable). Without alloc it returns false at the first nil embed.
 func fieldByIndex(v reflect.Value, index []int, alloc bool) (reflect.Value, bool) {
 	for _, i := range index {
 		if v.Kind() == reflect.Ptr {
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -200,9 +200,9 @@ func assign(dst reflect.Value, raw any) error {
 		}
 		return structFromMap(dst, props)
 	case reflect.Slice:
-		return assignSlice(dst, raw)
+		return assignSlice(dst, sv)
 	case reflect.Map:
-		return assignMap(dst, raw)
+		return assignMap(dst, sv)
 	default:
 		st := sv.Type()
 		dt := dst.Type()
@@ -221,10 +221,9 @@ func assign(dst reflect.Value, raw any) error {
 }
 
 // assignSlice builds a fresh slice so the mapped value never aliases the record.
-func assignSlice(dst reflect.Value, raw any) error {
-	sv := reflect.ValueOf(raw)
+func assignSlice(dst, sv reflect.Value) error {
 	if sv.Kind() != reflect.Slice {
-		return typeError(dst.Type(), raw)
+		return typeError(dst.Type(), sv.Interface())
 	}
 	out := reflect.MakeSlice(dst.Type(), sv.Len(), sv.Len())
 	for i := 0; i < sv.Len(); i++ {
@@ -237,17 +236,16 @@ func assignSlice(dst reflect.Value, raw any) error {
 }
 
 // assignMap builds a fresh map so the mapped value never aliases the record.
-func assignMap(dst reflect.Value, raw any) error {
-	sv := reflect.ValueOf(raw)
+func assignMap(dst, sv reflect.Value) error {
 	if sv.Kind() != reflect.Map {
-		return typeError(dst.Type(), raw)
+		return typeError(dst.Type(), sv.Interface())
 	}
 	dt := dst.Type()
 	out := reflect.MakeMapWithSize(dt, sv.Len())
 	for iter := sv.MapRange(); iter.Next(); {
 		key := iter.Key()
 		if !key.Type().AssignableTo(dt.Key()) {
-			return typeError(dt, raw)
+			return typeError(dt, sv.Interface())
 		}
 		val := reflect.New(dt.Elem()).Elem()
 		if err := assign(val, iter.Value().Interface()); err != nil {
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -187,9 +187,9 @@ func assign(dst reflect.Value, raw any) error {
 		dst = dst.Elem()
 	}
 	sv := reflect.ValueOf(raw)
-	dt := dst.Type()
 	switch dst.Kind() {
 	case reflect.Struct:
+		dt := dst.Type()
 		if sv.Type().AssignableTo(dt) {
 			dst.Set(sv) // driver-native struct, e.g. time.Time or a dbtype value
 			return nil
@@ -204,12 +204,14 @@ func assign(dst reflect.Value, raw any) error {
 	case reflect.Map:
 		return assignMap(dst, raw)
 	default:
+		st := sv.Type()
+		dt := dst.Type()
 		switch {
-		case sv.Type().AssignableTo(dt):
+		case st.AssignableTo(dt):
 			dst.Set(sv)
 		case isNumeric(dst.Kind()) && isNumeric(sv.Kind()):
 			return assignNumeric(dst, sv)
-		case sv.Kind() == dst.Kind() && sv.Type().ConvertibleTo(dt):
+		case sv.Kind() == dst.Kind() && st.ConvertibleTo(dt):
 			dst.Set(sv.Convert(dt)) // named type over the same primitive kind
 		default:
 			return typeError(dt, raw)
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -241,17 +241,16 @@ func assignMap(dst, sv reflect.Value) error {
 		return typeError(dst.Type(), sv.Interface())
 	}
 	dt := dst.Type()
+	if !sv.Type().Key().AssignableTo(dt.Key()) {
+		return typeError(dt, sv.Interface())
+	}
 	out := reflect.MakeMapWithSize(dt, sv.Len())
 	for iter := sv.MapRange(); iter.Next(); {
-		key := iter.Key()
-		if !key.Type().AssignableTo(dt.Key()) {
-			return typeError(dt, sv.Interface())
-		}
 		val := reflect.New(dt.Elem()).Elem()
 		if err := assign(val, iter.Value().Interface()); err != nil {
 			return err
 		}
-		out.SetMapIndex(key, val)
+		out.SetMapIndex(iter.Key(), val)
 	}
 	dst.Set(out)
 	return nil
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -22,6 +22,7 @@ package mapping
 
 import (
 	"fmt"
+	"math/bits"
 	"reflect"
 	"strings"
 	"sync"
@@ -257,8 +258,8 @@ func assignMap(dst, sv reflect.Value) error {
 	return nil
 }
 
-// assignNumeric converts between numeric kinds, erroring on overflow rather than
-// silently wrapping; a float is never coerced into an integer field.
+// assignNumeric converts between numeric kinds, returning an error rather than
+// silently losing range or precision. A float is never coerced into an integer.
 func assignNumeric(dst, src reflect.Value) error {
 	switch {
 	case src.CanInt():
@@ -275,6 +276,9 @@ func assignNumeric(dst, src reflect.Value) error {
 			}
 			dst.SetUint(uint64(n))
 		case dst.CanFloat():
+			if !intFitsFloat(n, dst.Type().Bits()) {
+				return precisionError(dst.Type(), n)
+			}
 			dst.SetFloat(float64(n))
 		}
 	case src.CanFloat():
@@ -285,11 +289,31 @@ func assignNumeric(dst, src reflect.Value) error {
 		if dst.OverflowFloat(f) {
 			return overflowError(dst.Type(), f)
 		}
+		if dst.Type().Bits() < src.Type().Bits() && float64(float32(f)) != f {
+			return precisionError(dst.Type(), f)
+		}
 		dst.SetFloat(f)
 	default:
 		return typeError(dst.Type(), src.Interface())
 	}
 	return nil
+}
+
+// intFitsFloat reports whether n is exactly representable in a float of the given
+// bit size; the mantissa holds 24 significant bits for float32, 53 for float64.
+func intFitsFloat(n int64, floatBits int) bool {
+	if n == 0 {
+		return true
+	}
+	u := uint64(n)
+	if n < 0 {
+		u = uint64(-n)
+	}
+	mantissa := 53
+	if floatBits == 32 {
+		mantissa = 24
+	}
+	return bits.Len64(u)-bits.TrailingZeros64(u) <= mantissa
 }
 
 func propsOf(raw any) (map[string]any, bool) {
@@ -309,6 +333,10 @@ func typeError(dst reflect.Type, raw any) error {
 
 func overflowError(dst reflect.Type, v any) error {
 	return fmt.Errorf("value %v is out of range for %s", v, dst)
+}
+
+func precisionError(dst reflect.Type, v any) error {
+	return fmt.Errorf("value %v cannot be represented exactly as %s", v, dst)
 }
 
 func isNumeric(k reflect.Kind) bool {
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -276,7 +276,7 @@ func assignNumeric(dst, src reflect.Value) error {
 			}
 			dst.SetUint(uint64(n))
 		case dst.CanFloat():
-			if !intFitsFloat(n, dst.Type().Bits()) {
+			if !intFitsFloat(n, dst.Type()) {
 				return precisionError(dst.Type(), n)
 			}
 			dst.SetFloat(float64(n))
@@ -299,19 +299,24 @@ func assignNumeric(dst, src reflect.Value) error {
 	return nil
 }
 
-// intFitsFloat reports whether n is exactly representable in a float of the given
-// bit size; the mantissa holds 24 significant bits for float32, 53 for float64.
-func intFitsFloat(n int64, floatBits int) bool {
+// intFitsFloat reports whether n is exactly representable in floatType. Known
+// widths use their mantissa bit count; any other falls back to an exact round-trip.
+func intFitsFloat(n int64, floatType reflect.Type) bool {
 	if n == 0 {
 		return true
+	}
+	var mantissa int
+	switch floatType.Bits() {
+	case 32:
+		mantissa = 24
+	case 64:
+		mantissa = 53
+	default:
+		return reflect.ValueOf(n).Convert(floatType).Convert(reflect.TypeFor[int64]()).Int() == n
 	}
 	u := uint64(n)
 	if n < 0 {
 		u = uint64(-n)
-	}
-	mantissa := 53
-	if floatBits == 32 {
-		mantissa = 24
 	}
 	return bits.Len64(u)-bits.TrailingZeros64(u) <= mantissa
 }
--- a/neo4j/internal/mapping/mapping.go
+++ b/neo4j/internal/mapping/mapping.go
@@ -15,12 +15,13 @@
  * limitations under the License.
  */
 
-// Package mapping converts a user struct into the map[string]any shape the bolt
-// layer expects for Cypher parameters. Nested struct values are returned as-is;
-// the bolt layer re-enters this package for each one.
+// Package mapping converts between user structs and the map[string]any shape the
+// bolt layer uses. StructAsMap flattens a struct into Cypher parameters;
+// MapToStruct is the reverse, filling a struct from a result map.
 package mapping
 
 import (
+	"fmt"
 	"reflect"
 	"strings"
 	"sync"
@@ -55,7 +56,7 @@ func StructAsMap(v any) (map[string]any, bool) {
 	fields := fieldsOf(rv.Type())
 	out := make(map[string]any, len(fields))
 	for _, f := range fields {
-		fv, ok := fieldByIndex(rv, f.index)
+		fv, ok := fieldByIndex(rv, f.index, false)
 		if !ok {
 			// Field is reachable only through a nil pointer embed.
 			continue
@@ -68,13 +69,16 @@ func StructAsMap(v any) (map[string]any, bool) {
 	return out, true
 }
 
-// fieldByIndex walks index from v, dereferencing pointer embeds. It reports
-// false if a nil pointer is hit mid-path.
-func fieldByIndex(v reflect.Value, index []int) (reflect.Value, bool) {
+// fieldByIndex walks index from v through pointer embeds. With alloc it allocates
+// nil embeds (false if one is unexported/unsettable); without, it stops false at the first nil.
+func fieldByIndex(v reflect.Value, index []int, alloc bool) (reflect.Value, bool) {
 	for _, i := range index {
 		if v.Kind() == reflect.Ptr {
 			if v.IsNil() {
-				return reflect.Value{}, false
+				if !alloc || !v.CanSet() {
+					return reflect.Value{}, false
+				}
+				v.Set(reflect.New(v.Type().Elem()))
 			}
 			v = v.Elem()
 		}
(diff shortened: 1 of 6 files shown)


neo4j/internal/mapping/mapping.go
```diff
@@ -68,13 +69,16 @@ func StructAsMap(v any) (map[string]any, bool) {
 	return out, true
 }
 
-// fieldByIndex walks index from v, dereferencing pointer embeds. It reports
-// false if a nil pointer is hit mid-path.
-func fieldByIndex(v reflect.Value, index []int) (reflect.Value, bool) {
+// fieldByIndex walks index from v through pointer embeds. With alloc it allocates
+// nil embeds (false if one is unexported/unsettable); without, it stops false at the first nil.
```
robsdedude:
```suggestion
// nil embeds (false if one is unexported/unsettable); without, it stops and returns false at the first nil.
```

Or such. Not sure what "stops false" is means otherwise 🤔 

Or maybe the whole description could do with a little rewording. I find it somewhat hard to understand.

neo4j/mapping_test.go
```diff
@@ -0,0 +1,189 @@
...
+		got, err := As[person](rec)
+		assertMapped(t, got, err, alice)
+	})
+	t.Run("single scalar column maps by column name", func(t *testing.T) {
+		t.Parallel()
+		rec := &db.Record{Keys: []string{"name"}, Values: []any{"Alice"}}
+		got, err := As[person](rec)
+		assertMapped(t, got, err, person{Name: "Alice"})
+	})
+	t.Run("pointer target is allocated", func(t *testing.T) {
```
robsdedude, line 70: Consider adding something along the lines of

```go
	t.Run("multi-hop pointer target is allocated", func(t *testing.T) {
		t.Parallel()
		rec := &db.Record{Keys: []string{"p"}, Values: []any{Node{Props: props}}}
		got, err := As[**person](rec)
		if err != nil {
			t.Fatalf("unexpected error: %v", err)
		}
		if got == nil || **got != alice {
			t.Fatalf("As[**person] = %#v, want %#v", got, &alice)
		}
	})
```

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+			case len(f.index) < len(winner.index):
+				winner, ambiguous = f, false // shallower: outright winner
+			case len(f.index) == len(winner.index):
+				ambiguous = true // same depth: ambiguous
+			}
+		}
+		if !ambiguous {
+			byName[name] = winner
+		}
+	}
```
robsdedude, line 377: This whole behavior is not very intuitive for me. I think the (public) docs should clearly state how ambiguity is handled.

Further, I'm not sure that having a silent failure mode like this (where ambiguity leads to the driver silently ignoring the fields altogether) is great DX...

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+			switch {
+			case len(f.index) < len(winner.index):
+				winner, ambiguous = f, false // shallower: outright winner
+			case len(f.index) == len(winner.index):
+				ambiguous = true // same depth: ambiguous
+			}
+		}
+		if !ambiguous {
+			byName[name] = winner
+		}
```
robsdedude, line 376: The ambiguous case does not seem to have any test coverage. Appending `else { panic("Ambiguity!") }` doesn't make any unit tests fail. I think it should.

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+func assign(dst reflect.Value, raw any) error {
+	if raw == nil {
+		return nil
+	}
+	if dst.Kind() == reflect.Ptr {
+		if dst.IsNil() {
+			dst.Set(reflect.New(dst.Type().Elem()))
+		}
+		return assign(dst.Elem(), raw)
+	}
```
robsdedude:
```suggestion
	for dst.Kind() == reflect.Ptr {
		if dst.IsNil() {
			dst.Set(reflect.New(dst.Type().Elem()))
		}
		dst = dst.Elem()
	}
```
Maybe I'm overlooking something, but this (copied from `MapToStruct`) seems equivalent except that is saves some recursion-depth and `nil` checks.

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+	sv := reflect.ValueOf(raw)
+	dt := dst.Type()
+	switch dst.Kind() {
+	case reflect.Struct:
+		if sv.Type().AssignableTo(dt) {
```
robsdedude, line 195: Quick and dirty test to show an issue with this

```go
	type withNode struct {
		Inner dbtype.Node
	}
	
	t.Run("Copies node properties", func(t *testing.T) {
		t.Parallel()

		node := dbtype.Node{Id: 1, ElementId: "1", Labels: []string{"Movie"}, Props: map[string]any{"Title": "FromNode", "Released": int64(3)}}
		src := map[string]any{"Inner": node}
		want := &withNode{Inner: node}

		got := reflect.New(reflect.TypeOf(want).Elem()).Interface()
		if err := MapToStruct(src, got); err != nil {
			t.Fatalf("unexpected error: %v", err)
		}
		if !reflect.DeepEqual(got, want) {
			t.Fatalf("MapToStruct mismatch\nwant: %#v\n got: %#v", want, got)
		}

		got.(*withNode).Inner.Props["Title"] = "Changed"
		if node.Props["Title"] != "FromNode" {
			t.Fatalf("MapToStruct did not copy node properties, got: %#v, want: %#v", got.(*withNode).Inner.Props["Title"], node.Props["Title"])
		}
	})
```

The issue here is that this is inconsistent with maps and slices where the mapping logic goes  out of its way to make sure the values are copied. For driver structs (including node properties, which also are a map), values aren't copied. I wonder if this is worth adjusting. If yes, all types that have exported fields would need to be copied to avoid this kind of indirect mutability (e.g., one could imagine `Duration.Nanos` being mutated).

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+			dst.Set(sv.Convert(dt)) // named type over the same primitive kind
+		default:
+			return typeError(dt, raw)
+		}
+		return nil
+	}
+}
+
+// assignSlice builds a fresh slice so the mapped value never aliases the record.
+func assignSlice(dst reflect.Value, raw any) error {
```
robsdedude: Could save a tiny bit of conversion work by making `assignSlice` take `raw reflect.Value` instead of `raw any`. The only call site is in `assign` and there we already have created `reflect.ValueOf(raw)`.

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+func assignMap(dst reflect.Value, raw any) error {
+	sv := reflect.ValueOf(raw)
+	if sv.Kind() != reflect.Map {
+		return typeError(dst.Type(), raw)
+	}
+	dt := dst.Type()
+	out := reflect.MakeMapWithSize(dt, sv.Len())
+	for iter := sv.MapRange(); iter.Next(); {
+		key := iter.Key()
+		if !key.Type().AssignableTo(dt.Key()) {
```
robsdedude: Commenting out the `AssignableTo` check makes no test fail => hole in the test suite.

Test sketches to be added to `TestMapToStructErrors`:

```go
	t.Run("TODO: name; withMapWrongKey", func(t *testing.T) {
		t.Parallel()
		var w struct {
			Scores map[int]int
		}

		err := MapToStruct(map[string]any{"Scores": map[string]any{"key": 123}}, &w)
		if err == nil {
			t.Fatal("expected error for withMapWrongKey")
		}
		if got := err.Error(); !strings.Contains(got, "\"Scores\"") || !strings.Contains(got, "map[string]interface") || !strings.Contains(got, "map[int]int") {
			t.Fatalf("error should mention field name and types involved, got: %q", got)
		}
	})
	t.Run("TODO: name; withMapWrongValue", func(t *testing.T) {
		t.Parallel()
		var w struct {
			Scores map[string]string
		}

		err := MapToStruct(map[string]any{"Scores": map[string]any{"key": 123}}, &w)
		if err == nil {
			t.Fatal("expected error for withMapWrongValue")
		}
		if got := err.Error(); !strings.Contains(got, "\"Scores\"") || !strings.Contains(got, "int") || !strings.Contains(got, "string") {
			t.Fatalf("error should mention field name and types involved, got: %q", got)
		}
	})
```

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+			return typeError(dt, raw)
+		}
```
robsdedude: Unless I'm missing something here, I think it's sufficient to to the key type check once. Both maps `dst` and `raw` have a single type for their keys; neither of which will change during iteration.

```suggestion
	if !sv.Type().Key().AssignableTo(dt.Key()) {
		return typeError(dt, raw)
	}
	out := reflect.MakeMapWithSize(dt, sv.Len())
	for iter := sv.MapRange(); iter.Next(); {
		key := iter.Key()
```

neo4j/internal/mapping/mapping_test.go
```diff
@@ -197,6 +198,417 @@ func TestStructAsMap(t *testing.T) {
...
+		{
+			name: "named type over string is converted",
+			src:  map[string]any{"State": "active"},
+			want: &withNamed{State: status("active")},
+		},
+		{
+			name: "typed map converts its values",
+			src:  map[string]any{"Scores": map[string]any{"a": int64(1), "b": int64(2)}},
+			want: &withTypedMap{Scores: map[string]int{"a": 1, "b": 2}},
+		},
```
robsdedude, line 357: Some more tests I find worthwhile adding

```suggestion
		},
		{
			name: "typed map converts its keys",
			src:  map[string]any{"Scores": map[string]any{"a": int64(1), "b": int64(2)}},
			want: &withMapAnyKey{Scores: map[any]int{"a": 1, "b": 2}},
		},
		{
			name: "typed map converts its keys and values",
			src:  map[string]any{"Scores": map[string]any{"a": int64(1), "b": int64(2)}},
			want: &withMapUntyped{Scores: map[any]any{"a": int64(1), "b": int64(2)}},
		},		
```

needs 

```go
	type withMapAnyKey struct {
		Scores map[any]int
	}
	type withMapUntyped struct {
		Scores map[any]any
	}
```

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+			dst.Set(sv)
+		case isNumeric(dst.Kind()) && isNumeric(sv.Kind()):
+			return assignNumeric(dst, sv)
+		case sv.Kind() == dst.Kind() && sv.Type().ConvertibleTo(dt):
+			dst.Set(sv.Convert(dt)) // named type over the same primitive kind
+		default:
+			return typeError(dt, raw)
+		}
+		return nil
+	}
```
robsdedude, line 222: Reduction of potentially semi-expensive `.Type()` calls.

```suggestion
	sv := reflect.ValueOf(raw)
	switch dst.Kind() {
	case reflect.Struct:
		dt := dst.Type()
		if sv.Type().AssignableTo(dt) {
			dst.Set(sv) // driver-native struct, e.g. time.Time or a dbtype value
			return nil
		}
		props, ok := propsOf(raw)
		if !ok {
			return typeError(dt, raw)
		}
		return structFromMap(dst, props)
	case reflect.Slice:
		return assignSlice(dst, raw)
	case reflect.Map:
		return assignMap(dst, raw)
	default:
		st := sv.Type()
		dt := dst.Type()
		switch {
		case st.AssignableTo(dt):
			dst.Set(sv)
		case isNumeric(dst.Kind()) && isNumeric(sv.Kind()):
			return assignNumeric(dst, sv)
		case sv.Kind() == dst.Kind() && st.ConvertibleTo(dt):
			dst.Set(sv.Convert(dt)) // named type over the same primitive kind
		default:
			return typeError(dt, raw)
		}
		return nil
	}
```

neo4j/internal/mapping/mapping_test.go
```diff
@@ -197,6 +198,417 @@ func TestStructAsMap(t *testing.T) {
...
+		t.Parallel()
+		var w struct{ Inner movie }
+		err := MapToStruct(map[string]any{"Inner": map[string]any{"Released": "nope"}}, &w)
+		if err == nil {
+			t.Fatal("expected error for bad nested property")
+		}
+		if got := err.Error(); !strings.Contains(got, "Inner") || !strings.Contains(got, "Released") {
+			t.Fatalf("error should name the property path, got: %q", got)
+		}
+	})
```
robsdedude, line 610: Suggesting some more tests
```suggestion
	})
	t.Run("does not coerce int64 into pointer", func(t *testing.T) {
		t.Parallel()
		var w struct {
			Foo uintptr
		}

		err := MapToStruct(map[string]any{"Foo": 123456}, &w)
		if err == nil {
			t.Fatal("expected error for uintptr field with int64 value")
		}
		if got := err.Error(); !strings.Contains(got, "\"Foo\"") || !strings.Contains(got, "int") || !strings.Contains(got, "uintptr") {
			t.Fatalf("error should mention field name and types involved, got: %q", got)
		}
	})	
```

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +131,226 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+			return err
+		}
+		out.SetMapIndex(key, val)
+	}
+	dst.Set(out)
+	return nil
+}
+
+// assignNumeric converts between numeric kinds, erroring on overflow rather than
+// silently wrapping; a float is never coerced into an integer field.
```
robsdedude: > a float is never coerced into an integer field.

That's true. But the other way around it is. And note that this can be lossy. Try

```go
	v := 9223372036854775807
	fmt.Println(v)
	fmt.Println(float64(v))
	fmt.Println(float64(v) - 9223372036854775000.0) // 807 if it were lossless, is 1024
	fmt.Println(int64(float64(v)))
```

Maybe this feature should be revised (I'm thinking either removing it, or at least adding a loss check). If the feature stays (in whichever shape), this coercion should be clearly documented in the *public* API docs.

Arguably, it's the same with assigning `float64` values to `float32` variables, which the logic does too. There only is a max and min check, but no guarantee that the resolution of the destination float type at the given value is granular enough to precisely represent the source value. I think that too should be documented.

neo4j/internal/mapping/mapping_test.go
```diff
@@ -197,6 +198,417 @@ func TestStructAsMap(t *testing.T) {
...
+		{
+			name: "scalar pointer field is allocated",
+			src:  map[string]any{"Name": "Trinity"},
+			want: &withScalarPointer{Name: &name},
+		},
+		{
+			name: "scalar slice",
+			src:  map[string]any{"Tags": []any{"a", "b"}},
+			want: &withSlice{Tags: []string{"a", "b"}},
+		},
```
robsdedude, line 387: Might be worth testing some slice inner type coercion:

```suggestion
		},
		{
			name: "untyped slice field coerces into typed slice field",
			src:  map[string]any{"Tags": []any{"a", "b"}},
			want: &withSlice{Tags: []string{"a", "b"}},
		},
		{
			name: "string slice coerces into untyped slice field",
			src:  map[string]any{"Things": []string{"a", "b"}},
			want: &withSliceUntyped{Things: []any{"a", "b"}},
		},
		{
			name: "mixed slice is assigned to untyped slice field",
			src:  map[string]any{"Things": []any{"a", "b", int64(123)}},
			want: &withSliceUntyped{Things: []any{"a", "b", int64(123)}},
		},		
```
with

```go
	type withSliceUntyped struct {
		Things []any
	}
```

N.B.: admittedly, case 2 never happens in the driver code as it stands today, I believe. The driver would only ever produce `[]any` slices. But this already works and asserting that it keeps working makes the mapping code more robust to possible future driver changes at pretty much no cost.

neo4j/internal/mapping/mapping_test.go
```diff
@@ -197,6 +198,417 @@ func TestStructAsMap(t *testing.T) {
...
+	type tagged struct {
+		Title string `neo4j:"title"`
+		Skip  string `neo4j:"-"`
+	}
+	type withNumeric struct {
+		Count   int
+		Rating  float64
+		Rating2 float32
+		Small   int32
+		Unsent  uint
```
robsdedude, line 221: Might be worth to test all number types supported (cf. `isNumeric`).

neo4j/internal/mapping/mapping_test.go
```diff
@@ -197,6 +198,417 @@ func TestStructAsMap(t *testing.T) {
...
+		{
+			name: "anonymous embedded struct is populated",
+			src:  map[string]any{"ID": "1", "Name": "Alice"},
+			want: &embedded{base: base{ID: "1"}, Name: "Alice"},
+		},
+		{
+			name: "nil pointer embed is allocated",
+			src:  map[string]any{"ID": "1", "Name": "Alice"},
+			want: &pointerEmbedded{Base: &Base{ID: "1"}, Name: "Alice"},
+		},
```
robsdedude, line 437: Maybe some more assertions about `nil` values.
```suggestion
		},
		{
			name: "nil pointer field is nil",
			src:  map[string]any{"Inner": nil},
			want: &withPointer{Inner: nil},
		},
		{
			name: "scalar pointer nil field is nil",
			src:  map[string]any{"Name": nil},
			want: &withScalarPointer{Name: nil},
		},		
```
Both `Inner: nil` and `Name: nil` are optional as they're zero values for pointers. As a non-native speaker of Go 😜 it took me a while to realize the implication for the mapping behavior for pointer properties as the docs that just state

> Missing or null properties leave zero values and extra properties are ignored.

neo4j/internal/mapping/mapping_test.go
```diff
@@ -197,6 +198,417 @@ func TestStructAsMap(t *testing.T) {
...
+		t.Parallel()
+		var s string
+		if err := MapToStruct(map[string]any{}, &s); err == nil {
+			t.Fatal("expected error for pointer to non-struct")
+		}
+	})
+	t.Run("type mismatch", func(t *testing.T) {
+		t.Parallel()
+		var m movie
+		if err := MapToStruct(map[string]any{"Released": "not-a-number"}, &m); err == nil {
```
robsdedude:
```suggestion
		if err := MapToStruct(map[string]any{"Released": "1999"}, &m); err == nil {
```

Is it maybe worth adding a test that we don't end up parsing integers or such?

neo4j/internal/mapping/mapping.go
robsdedude: Maybe the test case I suggest would fit better next to `TestMapToStructDoesNotAliasSource`.

neo4j/test-integration/mapping_test.go
```diff
@@ -0,0 +1,156 @@
...
+		assertEquals(t, people[0], omPerson{Name: "Alice", Age: 30})
+	})
+
+	outer.Run("time.Time property round-trips", func(t *testing.T) {
+		type event struct {
+			Name string    `neo4j:"name"`
+			At   time.Time `neo4j:"at"`
+		}
+		cleanup(t)
+		at := time.Date(2026, 5, 14, 10, 0, 0, 0, time.UTC)
```
robsdedude, line 118: Nice test, but UTC is boring. Consider adding a test with something spicy. A time zone with DST and a time that exists twice, for instance, once before and once after the offset shift.

neo4j/test-integration/mapping_test.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+type omPerson struct {
+	Name string `neo4j:"name"`
+	Age  int64  `neo4j:"age"`
+}
+
+// TestObjectMapping round-trips through a real server: nodes are created from a
+// struct via the input mapping and read back through the output mapping.
+func TestObjectMapping(outer *testing.T) {
```
robsdedude, line 36: Only testing mapping from Dictionaries and Nodes. Might be worth to add a more complex example using nodes and relationships: properly mapping a small graph.

neo4j/mapping.go
```diff
@@ -0,0 +1,92 @@
...
+	out := make([]T, len(records))
+	for i, record := range records {
+		v, err := As[T](record)
+		if err != nil {
+			return nil, err
+		}
+		out[i] = v
+	}
+	return out, nil
+}
```
robsdedude, line 87:
```suggestion
func CollectRecordsAs[T any](records []*Record) ([]T, error) {
	out := make([]T, 0, len(records))
	for _, record := range records {
		v, err := As[T](record)
		if err != nil {
			return nil, err
		}
		out = append(out, v)
	}
	return out, nil
}
```
I've not benchmarked it, but maybe this is slightly faster as it's not initializing 2 `T` values for each record (1 inside `As`, and 1 in `make` inside this function), but only 1 (inside `As`).

neo4j/mapping_example_test.go
```diff
@@ -0,0 +1,112 @@
...
+	defer handleClose(ctx, driver)
+	session := driver.NewSession(ctx, neo4j.SessionConfig{})
+	defer handleClose(ctx, session)
+
+	type Movie struct {
+		Title    string `neo4j:"title"`
+		Released int64  `neo4j:"released"`
+	}
+
+	result, err := session.Run(ctx, "MATCH (m:Movie {title: $title}) RETURN m", map[string]any{"title": "The Matrix"})
```
robsdedude: Uses `session.Run` for a query that doesn't need an auto-commit tx. We don't want our users to do this, so we shouldn't write examples with it either :)

neo4j/internal/mapping/mapping.go
StephenCathcart: added a lossless-or-error check so int to float and float64 to float32 now error instead of silently rounding, and documented the numeric coercion rules on As 👍

neo4j/internal/mapping/mapping.go
StephenCathcart: yeah, it's inconsistent. i'm not sure yet how far the copy rabbit hole goes across the driver types, so i'll document the alias caveat for node and relationship properties now and take the fuller fix as a follow-up. will raise it on the object-mapping discussion (#701) for usage feedback when it's merged.

neo4j/internal/mapping/mapping.go
StephenCathcart: yeah it's a bit awkward. i mirrored encoding/json's field promotion here: an outer field shadows a deeper one, and a same-depth tie is dropped rather than erroring. i'll document that on As and add a test for the ambiguous case 👍

neo4j/test-integration/mapping_test.go
StephenCathcart: tried the ambiguous fold first but it doesn't round-trip on 4.4: the legacy datetime sends local seconds + zone id, so the two 02:30s collapse to one offset (Testkit caught it - 107). swapped to a summer/winter Berlin pair, still covers both DST offsets without needing the server's tz resolution 👍

neo4j/internal/mapping/mapping.go
robsdedude: Point taken. I think mirroring behavior of 1st party go libs is a good thing DX-wise as it makes for an idiomatic API. For better or for worse :zany_face:

neo4j/internal/mapping/mapping.go
StephenCathcart: tracked as DRIVERS-553 and captured on the object mapping discussion (#701) so the copy-vs-alias decision doesn't get lost 👍

neo4j/internal/mapping/mapping.go
```diff
@@ -285,13 +289,33 @@ func assignNumeric(dst, src reflect.Value) error {
...
+		return true
+	}
+	u := uint64(n)
+	if n < 0 {
+		u = uint64(-n)
+	}
+	mantissa := 53
+	if floatBits == 32 {
+		mantissa = 24
+	}
```
robsdedude: I'm not sure if I'm being paranoid here, but this code will break should go ever decide to introduce more float types (e.g., float16 or float128), which would probably be considered an additive change. Options I see:
 * add a fallback case that does something like `n == int64(floatType(n))`
 * add a fallback case that panics
 * make a guess and implement the other float precisions defined in IEEE 754
 * declare me paranoid and do nothing 😇

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +133,252 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
```
robsdedude: I'm not sure if I'm being paranoid here, but this code will break should go ever decide to introduce more float types (e.g., float16 or float128), which would probably be considered an additive change. Options I see:
 * add a fallback case that does something like `n == int64(floatType(n))`
 * add a fallback case that panics
 * make a guess and implement the other float precisions defined in IEEE 754
 * declare me paranoid and do nothing 😇 

Option 1 could look something like this:

```suggestion
			if !intFitsFloat(n, dst.Type()) {
				return precisionError(dst.Type(), n)
			}
			dst.SetFloat(float64(n))
		}
	case src.CanFloat():
		if !dst.CanFloat() {
			return typeError(dst.Type(), src.Interface())
		}
		f := src.Float()
		if dst.OverflowFloat(f) {
			return overflowError(dst.Type(), f)
		}
		if dst.Type().Bits() < src.Type().Bits() && float64(float32(f)) != f {
			return precisionError(dst.Type(), f)
		}
		dst.SetFloat(f)
	default:
		return typeError(dst.Type(), src.Interface())
	}
	return nil
}

// intFitsFloat reports whether n is exactly representable in a float of the given
// bit size; the mantissa holds 24 significant bits for float32, 53 for float64.
func intFitsFloat(n int64, floatType reflect.Type) bool {
	if n == 0 {
		return true
	}
	var mantissa int
	switch floatType.Bits() {
	case 64:
		mantissa = 53
	case 32:
		mantissa = 24
	default:
		int64Type := reflect.TypeFor[int64]()
		nOriginal := reflect.ValueOf(n)
		nroundTripped := reflect.New(int64Type).Elem()
		floatValue := reflect.New(floatType).Elem()
		floatValue.Set(nOriginal.Convert(floatType))
		nroundTripped.Set(floatValue.Convert(int64Type))
		return nroundTripped.Equal(nOriginal)
	}
	u := uint64(n)
	if n < 0 {
		u = uint64(-n)
	}
```

neo4j/internal/mapping/mapping.go
```diff
@@ -127,6 +133,252 @@ func buildFields(t reflect.Type, indexPath []int) []fieldInfo {
...
+	if n == 0 {
+		return true
+	}
+	u := uint64(n)
+	if n < 0 {
+		u = uint64(-n)
+	}
+	mantissa := 53
+	if floatBits == 32 {
+		mantissa = 24
```
robsdedude: Nit: I'd appreciate more tests

Changing the mantissa for the 32-bit case to something silly like `1` makes no test fail 👀 There should probably be tests that fail if it's changed to anything but 24 (especially interesting 23 and 25). Likewise is the 64-bit case rather lenient. You have to change the number by a lot to make any test fail.

neo4j/mapping.go
```diff
@@ -0,0 +1,101 @@
...
+// neo4j struct tags: `neo4j:"name"` renames a field, `neo4j:"-"` skips it.
+// Missing or null properties leave zero values and extra properties are ignored.
+// A record with a single node, relationship, or map column maps that value's
+// properties; otherwise its columns map by name.
+//
+// Numeric properties are converted to the field's type; a conversion that would
+// lose range or precision returns an error instead of silently truncating.
+//
+// Slice and map fields are copied; a node or relationship's properties are not,
+// so a mapped value may share that data with the record.
```
robsdedude: Nit
```suggestion
// Slice and map fields are recursively copied; a node or relationship's
// properties are not, so a mapped value may share that data with the record..
```
Or maybe "deeply copied" or such.

robsdedude (approved):
Only small suggestions left. No blockers.

neo4j/internal/mapping/mapping.go
StephenCathcart: added boundary tests at the mantissa edge (2^24-1 fits, 2^24+1 doesn't; same at 2^53 for float64) so shifting the mantissa off 24/53 now fails 👍

neo4j/internal/mapping/mapping.go
StephenCathcart: went with option 1: for a float width that isn't 32 or 64 it falls back to an exact round-trip instead of guessing the mantissa 👍

neo4j/internal/mapping/mapping_test.go
```diff
@@ -197,6 +198,592 @@ func TestStructAsMap(t *testing.T) {
...
+		t.Parallel()
+		var w struct{ N float32 }
+		if err := MapToStruct(map[string]any{"N": int64(1<<24 - 1)}, &w); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		if w.N != 16777215 {
+			t.Fatalf("N = %v, want 16777215", w.N)
+		}
+	})
+	t.Run("largest exact integer for float64", func(t *testing.T) {
```
robsdedude, line 477: Cool! Maybe do negative as well 😍

PR #709, opened 2026-09-07
Port Pin pre-commit staticcheck to the version CI uses

StephenCathcart:
Port of https://github.com/neo4j/neo4j-go-driver/pull/708

Closes: DRIVERS-539


robsdedude (approved):
📍️
Note that 0.3.3 will be running in CI while 0.6.0 will be running during local development (unless developing with an old tool chain).

